PR #26, opened 2020-03-29
get causality returns causality

thorade:


chria:
Thanks for the correction.

PR #27, opened 2020-03-30
Point to conda-forge

giplessis:
Dear JModelica people,
I first try to install PyFMI using conda and following the instructions from
https://jmodelica.org/pyfmi/installation.html#
I run into the problem of the FMILibrarynot being installed but after a while understood that the new version is now hosted on conda-forge and not in https://conda.binstar.org/chria anymore.
Maybe this webpage could be updated to point to conda-forge.

Best regards,

### DIFF
--- a/doc/sphinx/source/installation.rst
+++ b/doc/sphinx/source/installation.rst
@@ -5,24 +5,30 @@ Installation
 =============
 
 
+Using CONDA
+-----------
+
+PyFMI is available through `conda <http://conda.pydata.org/docs/index.html>`_::
+
+    conda install -c conda-forge pyfmi
+    
+Using PIP
+---------
+
 Current version:
 
     Available on PyPI, http://pypi.python.org/pypi/PyFMI
 
-Additionally, PyFMI is available through `conda <http://conda.pydata.org/docs/index.html>`_::
-
-    conda install -c https://conda.binstar.org/chria pyfmi
-
 Requirements:
--------------
+^^^^^^^^^^^^^
 - `FMI Library <http://www.jmodelica.org/FMILibrary>`_
 - `Numpy <http://pypi.python.org/pypi/numpy>`_
 - `Scipy <http://pypi.python.org/pypi/scipy>`_
 - `lxml <http://pypi.python.org/pypi/lxml>`_
 - `Assimulo <http://pypi.python.org/pypi/Assimulo>`_
 
-Optional
----------
+Optional:
+^^^^^^^^^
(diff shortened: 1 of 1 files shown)


jschueller:
@modelonrobinandersson is there a reason not to integrate this ?

modelonrobinandersson:
> @modelonrobinandersson is there a reason not to integrate this ?

Hi! Let me get back to this later this week.

modelonrobinandersson:
Hi @jschueller, I tried installing PyFMI via conda as in the proposed changeset however I fail to run `pyfmi.examples.fmi_bouncing_ball`, the error I get is `ModuleNotFoundError: No module named 'pylab'`. Therefore I tried installing `matplotlib` to resolve the `pylab` issue but instead get the following error
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
pyfmi 2.9.8 requires cython>=0.29.13, which is not installed.
pyfmi 2.9.8 requires nose>=1.3.7, which is not installed.
```
Still I ignored this error and the pyfmi example works, however do you think it would be possible to fix the errors I faced? It would be helpful as I don't think everyone would be able to resolve this on their own.

I used a clean centos 7.4 environment when testing this (via docker).

modelonrobinandersson:
> I suggest to install matplotlib from the same channel too, can you retry with: `conda install -c conda-forge pyfmi matplotlib` If that doesnt work, try to install from miniforge as it enforces packages come from the same channel to avoid conflicts
It does work but downgrades a couple of packages:
```
Yes that does work

modelonrobinandersson:
Hi @jschueller do you know who is responsible for the uploads at conda-forge? I wonder if it is possible to change the home page URL for both `assimulo` and `pyfmi`.
I am planning to update the README for both to point at conda-forge moving forward.

jschueller:
yes, I can easily change it, what are the new links ?

modelonrobinandersson:
> yes, I can easily change it, what are the new links ?

I would propose to simply point to the respective GitHub pages, i.e. for PyFMI https://github.com/modelon-community/PyFMI and for Assimulo https://github.com/modelon-community/Assimulo.

Thank you in advance!

jschueller:
done, it will be updated on the next version

modelonrobinandersson:
Hi @giplessis @jschueller, closing this as https://github.com/modelon-community/PyFMI/pull/189 covers this one. I will merge it once the review is done.

PR #39, opened 2020-05-20
Dev ak 3189

modelonandrijakukolj:
OCT-3189 implementing serialization/de-serialization of saved FMU state


chria:
Looks good! I'll add a changelog to it as well

PR #70, opened 2020-12-23
Make sure option "logging" works for different solvers

agnesramle:
It should be possible to turn on simulate option "logging" for all different kinds of solvers without getting exceptions. The log file should display what is supported.


modelonrobinandersson:
Looks good @agnesramle!

PR #73, opened 2021-01-26
Fixed initial for ScalarVariable2

agnesramle:
Added so initial is set properly on the ScalarVariable2.


modelonrobinandersson (changes requested):
This looks clean and good. Good that the test is updated as well. Can you update the changelog when merging?

PR #75, opened 2021-02-08
Corrected get_variable_nominal for valueref

agnesramle:
get_variable_nominal on FMU 1.0 objects returned an exception previously.


modelonrobinandersson (approved):
Looks good, also great that you added the simple test for this

Issue #76, opened 2021-02-22
pyfmi.fmi.FMUException: Failed to retrieve the continuous states


loic-fejoz:
On [a FMI2.0 model generated by OpenModelica (SinusGenerator.fmu.zip)](https://github.com/modelon-community/PyFMI/files/6025333/SinusGenerator.fmu.zip), I have the following exception with version 2.8.5:
```
Traceback (most recent call last):
  File "./fmi-comgw-connector.py", line 12, in <module>
    res = model.simulate() #final_time=10)
  File "src/pyfmi/fmi.pyx", line 7873, in pyfmi.fmi.FMUModelME2.simulate
  File "src/pyfmi/fmi.pyx", line 319, in pyfmi.fmi.ModelBase._exec_simulate_algorithm
  File "src/pyfmi/fmi.pyx", line 313, in pyfmi.fmi.ModelBase._exec_simulate_algorithm
  File "/opt/conda/lib/python3.8/site-packages/pyfmi/fmi_algorithm_drivers.py", line 385, in __init__
    self.probl = FMIODE2(self.model, result_file_name=self.result_file_name, with_jacobian=self.with_jacobian, start_time=self.start_time,logging=self.options["logging"], result_handler=self.result_handler,extra_equations=self.options["extra_equations"])
  File "src/pyfmi/simulation/assimulo_interface.pyx", line 592, in pyfmi.simulation.assimulo_interface.FMIODE2.__init__
  File "src/pyfmi/fmi.pyx", line 7705, in pyfmi.fmi.FMUModelME2._get_continuous_states
pyfmi.fmi.FMUException: Failed to retrieve the continuous states.
make: *** [Makefile:11 : run] Erreur 1
```

The exact version of each package is:
```
#21 [16/16] RUN conda list
#21 sha256:0548617c3561e7b24e7586b50e78650f97787acbdc5d1b954ea4844d7efca80e
#21 1.205 # packages in environment at /opt/conda:
#21 1.205 #
#21 1.205 # Name                    Version                   Build  Channel
#21 1.205 _libgcc_mutex             0.1                 conda_forge    conda-forge
#21 1.205 _openmp_mutex             4.5                       1_gnu    conda-forge
#21 1.205 assimulo                  3.2.3            py38hd6244a5_3    conda-forge
#21 1.205 brotlipy                  0.7.0           py38h27cfd23_1003  
#21 1.205 ca-certificates           2020.12.5            ha878542_0    conda-forge
[...]

PR #78, opened 2021-04-01
Allow PARAMETER causality on ME FMUs

jschueller:
Currently pyfmi does not allow to simulate a model-exchange FMU when one decide to set the value of a variable of causality PARAMETER (works fine with CS though).

```
model ParamInput

  parameter Real p1=1.0;
  Real a;
  output Real out1;
  Real d;
equation
  a = p1 + 10;
  out1 = a * 2;
  der(d) = d + 2;
end ParamInput;
```

here is how to run it:
```
import pyfmi
model = pyfmi.load_fmu('ParamInput.fmu')
res = model.simulate(input=(['p1'], lambda t:2*t))
```

one dummy d variable was added with a differential equation to avoid issue #76 when simulating ME models, thanks @Claire-Eleutheriane for the trick

### DIFF
--- a/src/pyfmi/simulation/assimulo_interface.pyx
+++ b/src/pyfmi/simulation/assimulo_interface.pyx
@@ -663,8 +663,8 @@ cdef class FMIODE2(cExplicit_Problem):
                 input_names = [input_names]
                 
             for i,name in enumerate(input_names):
-                if self._model.get_variable_causality(name) != fmi.FMI2_INPUT:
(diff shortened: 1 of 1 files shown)


jschueller:
hello @chria,
I saw you just released 2.8.6,
any chance this get picked up by the next one ?

chria:
Hi! I looked at it and according to the FMI specification you are not allowed to change parameters after the FMU have been initialized. So in general this won't work (and should not work) I'm afraid.

jschueller:
Hello,
Thanks for the analysis, it seems I have to use model.set then.

PR #80, opened 2021-04-05
Updated options descriptions

chria:


src/pyfmi/master.pyx
```diff
@@ -272,11 +272,17 @@ class MasterAlgOptions(OptionBase):
...
+            True.
+            Default: 1e-6
+
         result_file_name --
             Specifies the name of the file where the simulation result is
-            written. Setting this option to an empty string results in a default
-            file name that is based on the name of the model class.
-            Default: Empty string
+            written. Note that there should be one name for each model.
+            Default: Model name + "_result." + filetype
```
agnesramle, line 285: It should be specified here what format result_file_name should be passed in as since there is one name for each model.

src/pyfmi/master.pyx
```diff
@@ -299,8 +305,41 @@ class MasterAlgOptions(OptionBase):
...
+            values are: "greedy", "simply", "grouping". Each has their
+            pros and cons, for best performance, please test.
+            Default: "greedy"
 
-
+        block_initialization_order --
+            If set, manually overrrides the order in which the connected
+            models should be initialized. This only has an effect when
+            the option 'block_initialization' is set to True.
+            Default: None
```
agnesramle, line 334: Also here it is unclear which format this should be specified in.

src/pyfmi/master.pyx
```diff
@@ -1157,10 +1196,19 @@ cdef class Master:
                 result_object = ResultHandlerDummy(model)
             else:
                 raise fmi.FMUException("Currently only writing result to file (txt and binary) and none is supported.")
+            if not isinstance(opts["result_file_name"], dict):
+                raise fmi.FMUException("The result file names needs to be stored in a dict with the individual models as key.")
             from pyfmi.fmi_algorithm_drivers import FMICSAlgOptions
             local_opts = FMICSAlgOptions()
             prefix = "txt" if opts["result_handling"] == "file" else "mat"
-            local_opts["result_file_name"] = model.get_identifier()+'_'+str(i)+'_result.'+prefix
+            print(opts["result_file_name"])
```
agnesramle, line 1204: Is this debug printout?

tests/test_fmi_master.py
```diff
@@ -157,6 +167,46 @@ def test_basic_simulation_mat_file(self):
...
+    def test_basic_simulation_txt_file_naming_exists(self):
+        models, connections = self._load_basic_simulation()
+        
+        opts = {"result_handling":"file", "result_file_name": {models[0]: "Test1.txt", models[1]: "Test2.txt"}}
+        
+        res = self._sim_basic_simulation(models, connections, opts)
+        
+        assert os.path.isfile("Test1.txt"), "Test1.txt does not exists"
+        assert os.path.isfile("Test2.txt"), "Test2.txt does not exists"
+
```
agnesramle, line 209: I miss tests where result_handling is set to binary but result_file_name are specified as *.txt and vice versa.

agnesramle (changes requested):
Please see comments before merging to master.

src/pyfmi/master.pyx
chria: Updated description.

src/pyfmi/master.pyx
chria: Looking at this again, I believe this should be an experimental option. I've removed the description and marked the option as experimental instead.

src/pyfmi/master.pyx
chria: Good catch! Removed.

tests/test_fmi_master.py
chria: There is actually no restrictions on the suffix when the user provides a filename for the result. It is ok to provide .txt for binary files and .mat for text files. So no need to add a test.

agnesramle (approved):
Great, this looks good now!

PR #82, opened 2021-04-18
On demand trajectory loading + Always consistent binary file

chria:
Added support for loading trajectories from binary files on demand instead of all at the same time. This has a large performance gain when there are large result files and only a few trajectories are interesting. Furthermore, added support when writing binary files so that they are always in state where they can be reloaded.

### DIFF
--- a/src/common/io.py
+++ b/src/common/io.py
@@ -23,6 +23,7 @@ import codecs
 import re
 import sys
 import io
+import os
 
 import numpy as N
 import numpy as np
@@ -1146,26 +1147,26 @@ class ResultDymolaBinary(ResultDymola):
             fname --
                 Name of file.
         """
-        self._fname = fname
-        self._is_filestream = isinstance(fname, io.IOBase)
-        
-        if delayed_trajectory_loading:
-            
-            if self._is_filestream:
-                delayed = DelayedVariableLoad(self._fname, chars_as_strings=False)
-                self.raw = delayed.get_variables(variable_names=["name", "dataInfo", "data_2"])
+        if isinstance(fname, io.IOBase):
+            if hasattr(fname, "name") and os.path.isfile(fname.name):
+                self._fname = fname.name
             else:
-                with open(fname, "rb") as f:
-                    delayed = DelayedVariableLoad(f, chars_as_strings=False)
-                    self.raw = delayed.get_variables(variable_names=["name", "dataInfo", "data_2"])
-            
+                raise JIOError("Not supported file format, needs to be a filename or a stream based on a file.")
+        else:
+            self._fname = fname
+
+        if delayed_trajectory_loading:
+            with open(self._fname, "rb") as f:
+                delayed = DelayedVariableLoad(f, chars_as_strings=False)
+                self.raw = delayed.get_variables(variable_names=["name", "dataInfo", "data_2"])
+        
             self._data_2_info = self.raw["data_2"]
             self._data_2 = {}
             self._name_info   = self.raw["name"]
             
             self.name_lookup = self._get_name_dict()
         else:
-            self.raw = scipy.io.loadmat(fname,chars_as_strings=False, variable_names=["name", "dataInfo", "data_2"])
+            self.raw = scipy.io.loadmat(self._fname,chars_as_strings=False, variable_names=["name", "dataInfo", "data_2"])
             self._data_2 = self.raw["data_2"]
             
             name = self.raw['name']
@@ -1197,10 +1198,7 @@ class ResultDymolaBinary(ResultDymola):
         max_length     = self._name_info["max_length"]
         nbr_variables  = self._name_info["nbr_variables"]
 
-        if self._is_filestream:
-            name_dict = fmi_util.read_name_list(self._fname.fileno(), file_position, sizeof_type, int(nbr_variables), int(max_length))
-        else:
-            name_dict = fmi_util.read_name_list(encode(self._fname), file_position, sizeof_type, int(nbr_variables), int(max_length))
+        name_dict = fmi_util.read_name_list(encode(self._fname), file_position, sizeof_type, int(nbr_variables), int(max_length))
         
         return name_dict
         
@@ -1214,10 +1212,7 @@ class ResultDymolaBinary(ResultDymola):
             nbr_points     = self._data_2_info["nbr_points"]
             nbr_variables  = self._data_2_info["nbr_variables"]
             
-            if self._is_filestream:
-                self._data_2[data_index] = fmi_util.read_trajectory(self._fname.fileno(), data_index, file_position, sizeof_type, int(nbr_points), int(nbr_variables))
-            else:
-                self._data_2[data_index] = fmi_util.read_trajectory(encode(self._fname), data_index, file_position, sizeof_type, int(nbr_points), int(nbr_variables))
+            self._data_2[data_index] = fmi_util.read_trajectory(encode(self._fname), data_index, file_position, sizeof_type, int(nbr_points), int(nbr_variables))
             
             return self._data_2[data_index]
         else:
--- a/src/common/io.py
+++ b/src/common/io.py
@@ -33,6 +33,9 @@ import pyfmi.fmi as fmi
 import pyfmi.fmi_util as fmi_util
 from pyfmi.common import python3_flag, encode, decode
 
+from scipy.io.matlab.mio4 import MatFile4Reader, VarReader4, convert_dtypes, mdtypes_template
+import struct
+
 SYS_LITTLE_ENDIAN = sys.byteorder == 'little'
 
 class Trajectory:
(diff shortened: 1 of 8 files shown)


src/pyfmi/fmi_util.pyx
```diff
@@ -1000,99 +1000,81 @@ cdef class DumpData:
...
+cdef extern from "stdio.h":
+    FILE *fdopen(int, const char *)
+
+DTYPE = np.double
+ctypedef np.double_t DTYPE_t
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def read_trajectory(file_name, int data_index, int file_position, int sizeof_type, int nbr_points, int nbr_variables):
+    cdef int i = 0
```
agnesramle, line 1016: I miss a docstring describing the function and the arguments.

src/pyfmi/fmi_util.pyx
```diff
@@ -1000,99 +1000,81 @@ cdef class DumpData:
...
-                self.data[k] = i[k-self.rs]
-            for k in range(self.ris, self.ribs):
-                self.data[k] = b[k-self.ris]
-            
-            return self._create_string(self.data)
+    return data
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def read_name_list(file_name, int file_position, int sizeof_type, int nbr_variables, int max_length):
```
agnesramle, line 1043: I miss a docstring describing the function and the arguments.

src/pyfmi/fmi_util.pyx
```diff
@@ -1000,99 +1000,81 @@ cdef class DumpData:
...
-    cpdef _join_string(self, int used_length_cache):
-        cdef int j
+    fseek(cfile, file_position, 0)
+    for i in range(nbr_variables):
+        fread(<void*>(tmp), max_length, 1, cfile)
+        py_str = tmp
         
-        return "%.14E "%self.model.time + ' '.join(self.cache[:used_length_cache])
-
+        if i == 0:
```
agnesramle, line 1062: Why should this only hold true for the first variable read?

agnesramle (changes requested):

tests/test_io.py
```diff
@@ -509,7 +509,14 @@ def test_read_all_variables(self):
         for var in res.name:
             res.get_variable_data(var)
     
-    
+    @testattr(stddist = True)
+    def test_read_all_variables_from_stream(self):
```
agnesramle, line 513: This test does not ensure that res.name actually contains any names and that get_variable_data returns something meaningful. One option is to compare loading without a stream.

src/common/io.py
```diff
@@ -1146,26 +1147,26 @@ def __init__(self, fname, delayed_trajectory_loading = True):
...
                 Name of file.
         """
-        self._fname = fname
-        self._is_filestream = isinstance(fname, io.IOBase)
-        
+        if isinstance(fname, io.IOBase):
+            if hasattr(fname, "name") and os.path.isfile(fname.name):
+                self._fname = fname.name
+            else:
+                raise JIOError("Not supported file format, needs to be a filename or a stream based on a file.")
```
agnesramle, line 1159: Why is this limitation required?

src/common/io.py
```diff
@@ -1146,11 +1147,17 @@ def __init__(self, fname, delayed_trajectory_loading = True):
                 Name of file.
         """
         self._fname = fname
+        self._is_filestream = isinstance(fname, io.IOBase)
```
agnesramle: Docstring to init should be updated with that this is possible as well.

agnesramle (changes requested):

src/common/io.py
```diff
@@ -2072,11 +2072,14 @@ def _data_header(self, name, nbr_rows, nbr_cols, data_type):
         
         return header
         
+    def __write_header(self, header, name):
+        self._file.write(header.tostring(order="F"))
```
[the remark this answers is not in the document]
chria: Fixed

tests/test_io.py
```diff
@@ -556,6 +556,36 @@ def test_work_flow_me2(self):
...
+        derh = res.get_variable_data('der(h)')
+        g = res.get_variable_data('g')
+
+        nose.tools.assert_almost_equal(h.x[0], 1.000000, 5)
+        nose.tools.assert_almost_equal(derh.x[0], 0.000000, 5)
+        nose.tools.assert_almost_equal(h.x[1], 1.000000, 5)
+        nose.tools.assert_almost_equal(derh.x[1], 0.000000, 5)
+        nose.tools.assert_almost_equal(h.x[2], 1.000000, 5)
+        nose.tools.assert_almost_equal(derh.x[2], 0.000000, 5)
+
```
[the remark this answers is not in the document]
chria: Added error message.

src/common/io.py
```diff
@@ -1106,13 +1106,37 @@ def append(self, res):
...
 
+
+from scipy.io.matlab.mio4 import MatFile4Reader, VarReader4, convert_dtypes, mdtypes_template
+import struct
+
+#Overriding SCIPYs default reader for MATLAB v4 format
+class DelayedVarReader4(VarReader4):
+    def read_sub_array(self, hdr, copy=True):
+        if hdr.name == b"data_2":
+            return {"section": "data_2", "file_position": self.mat_stream.tell(), "sizeof_type": hdr.dtype.itemsize, "nbr_points": hdr.dims[1], "nbr_variables": hdr.dims[0]}
```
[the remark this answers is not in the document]

src/common/io.py
```diff
@@ -1106,13 +1106,37 @@ def append(self, res):
...
+        elif hdr.name == b"name":
+            return {"section": "name", "file_position": self.mat_stream.tell(), "sizeof_type": hdr.dtype.itemsize, "max_length": hdr.dims[0], "nbr_variables": hdr.dims[1]}
```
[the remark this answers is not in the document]

src/common/io.py
```diff
@@ -1106,13 +1106,37 @@ def append(self, res):
         self.data[1] = N.vstack((self.data[1],res.data[1]))
         self.data[1][n_points:,0] = self.data[1][n_points:,0] + time_shift 
 
+
+from scipy.io.matlab.mio4 import MatFile4Reader, VarReader4, convert_dtypes, mdtypes_template
+import struct
+
```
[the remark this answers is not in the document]
chria: Moved up.

src/common/io.py
```diff
@@ -1106,13 +1106,37 @@ def append(self, res):
...
 class ResultDymolaBinary(ResultDymola):
     """ 
     Class representing a simulation or optimization result loaded from a Dymola 
     binary file.
     """
 
-    def __init__(self,fname):
+    def __init__(self, fname, delayed_trajectory_loading = True):
         """
```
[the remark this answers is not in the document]
chria: Documented.

src/common/io.py
```diff
@@ -1122,16 +1146,70 @@ def __init__(self,fname):
...
-        self.name_lookup = {key:ind for ind,key in enumerate(self._name)}
         
+        self._delayed_loading = delayed_trajectory_loading
         self._description = None  
+        self._data_1      = None
+        
+    def _get_data_1(self):
+        if self._data_1 is None:
+            data_1 = scipy.io.loadmat(self._fname,chars_as_strings=False, variable_names=["data_1"])["data_1"]
+            self._data_1 = data_1
```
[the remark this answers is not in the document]

src/pyfmi/fmi_util.pyx
chria: Fixed.

src/pyfmi/fmi_util.pyx
chria: Fixed.

src/common/io.py
```diff
@@ -1146,11 +1147,17 @@ def __init__(self, fname, delayed_trajectory_loading = True):
...
-            with open(fname, "rb") as f:
-                delayed = DelayedVariableLoad(f, chars_as_strings=False)
+            
+            if self._is_filestream:
+                delayed = DelayedVariableLoad(self._fname, chars_as_strings=False)
                 self.raw = delayed.get_variables(variable_names=["name", "dataInfo", "data_2"])
+            else:
+                with open(fname, "rb") as f:
+                    delayed = DelayedVariableLoad(f, chars_as_strings=False)
+                    self.raw = delayed.get_variables(variable_names=["name", "dataInfo", "data_2"])
```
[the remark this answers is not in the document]

src/pyfmi/fmi_util.pyx
chria: It can hold true for others but if it does it means that the length is less than four. And if so then there is not problem as it won't cause any large overhead.

tests/test_io.py
chria: Fixed.

src/common/io.py

src/common/io.py
chria: The underlying Cython implemention requires a file in order to be efficient so for now, we require this.

src/pyfmi/fmi_util.pyx
agnesramle: I think it would be worth to have a comment on this in the code.

PR #88, opened 2021-04-23
Fixed segfault when storing large binary results

chria:
Fixed issue when trying to store results for a large number of variables. Added checks if allocations where successful or not and also added a specific method when description is not asked for.


agnesramle (commented):
LGTM! What about a test case?

chria:
@agnesramle  Added test.

agnesramle:
LGTM!

PR #90, opened 2021-04-27
Loading of unzipped FMUs

modelonrobinandersson:
This PR contains updates in order to load already unzipped FMUs


src/pyfmi/fmi.pyx
```diff
@@ -1190,13 +1190,14 @@ cdef class FMUModelBase(ModelBase):
     """
     An FMI Model loaded from a DLL.
     """
-    def __init__(self, fmu, path='.', enable_logging=None, log_file_name="", log_level=FMI_DEFAULT_LOG_LEVEL, _unzipped_dir=None, _connect_dll=True):
+    def __init__(self, fmu, path='.', enable_logging=None, log_file_name="", log_level=FMI_DEFAULT_LOG_LEVEL,
+                 _unzipped_dir=None, _connect_dll=True, allow_unzipped_fmu = False):
         """
```
chria, line 1195: We are missing a good docstring on this class, lets take the opportunity to add it.

src/pyfmi/fmi.pyx
```diff
@@ -1213,15 +1214,19 @@ cdef class FMUModelBase(ModelBase):
...
         self.callbacks.malloc  = FMIL.malloc
         self.callbacks.calloc  = FMIL.calloc
         self.callbacks.realloc = FMIL.realloc
         self.callbacks.free    = FMIL.free
         self.callbacks.logger  = importlogger
         self.callbacks.context = <void*>self #Class loggger
-        
-        
+
+        display_path_deprecation(path)
```
chria, line 1228: I'm a bit unsure about the gain of adding an extra method just to display the message. I'd rather see the message directly here instead of invoking an extra method (and also even though this message will be shown in a few places)

src/pyfmi/fmi.pyx
```diff
@@ -8222,12 +8261,49 @@ cdef class FMUModel(FMUModelME1):
...
+            raise FMUException(err_msg)
+    else:
+        # Check that the file referenced by fmu is appropriate
+        if not fmu_full_path.endswith('.fmu' if isinstance(fmu_full_path, str) else encode('.fmu')):
+            raise FMUException("FMUModel must be instantiated with an FMU (.fmu) file.")
+
+        if not os.path.isfile(fmu_full_path):
+            raise FMUException('Could not locate the FMU in the specified directory.')
+
+cdef FMIL.fmi_version_enu_t import_and_get_version(FMIL.fmi_import_context_t* context, char* fmu_full_path, char* fmu_temp_dir, int allow_unzipped_fmu):
```
chria, line 8296: I believe we should remove this method and directly invoke fmi_import_and_get_version. The only difference here is that we send in NULL in one branch. This can be handled directly at the places where this class is invoked.

src/pyfmi/fmil_import.pxd
```diff
@@ -1017,5 +1017,3 @@ cdef extern from 'fmilib.h':
```
chria, line 1017: Should there not be an extra method referenced here? (the import_and_get...)

src/pyfmi/fmi.pyx
```diff
@@ -8222,12 +8261,49 @@ cdef class FMUModel(FMUModelME1):
...
+            msg = "Argument named 'fmu' must be a directory if argument 'allow_unzipped_fmu' is set to True."
+            raise FMUException(msg)
+        if not os.path.isfile(os.path.join(fmu, 'modelDescription.xml')):
+            err_msg = "Specified fmu path '{}' needs".format(fmu)
+            err_msg += " to contain a modelDescription.xml according to the FMI specification."
+            raise FMUException(err_msg)
+    else:
+        # Check that the file referenced by fmu is appropriate
+        if not fmu_full_path.endswith('.fmu' if isinstance(fmu_full_path, str) else encode('.fmu')):
+            raise FMUException("FMUModel must be instantiated with an FMU (.fmu) file.")
```
chria, line 8291: Can we replace FMUModel with some thing else? Or someone change the wording so that it better fits all classes?

tests/test_fmi.py
```diff
@@ -36,6 +39,16 @@
...
 file_path = os.path.dirname(os.path.abspath(__file__))
 
+
+def _helper_unzipped_fmu_exception_invalid_dir(fmu_loader):
+    """ Verify that we get an exception if unzipped FMU does not contain modelDescription.xml, which it should according to the FMI specification.
+        The input argument is any of the FMU interfaces FMUModelME1, FMUModelME2, FMUModelCS1, FMUModelCS2 and load_fmu from pyfmi.fmi.
+    """
+    err_msg = "Specified fmu path '.*\\' needs to contain a modelDescription.xml according to the FMI specification"
+    with tempfile.TemporaryDirectory() as temp_dir:
+        with np.testing.assert_raises_regex(FMUException, err_msg):
```
chria, line 49: Why a with statement?

tests/test_fmi.py
```diff
@@ -91,6 +104,35 @@ def get_result(self):
         
 
 class Test_FMUModelME1:
+
+    @testattr(stddist = True)
+    def test_unzipped_fmu_exception_invalid_dir(self):
+        """ Verify that we get an exception if unzipped FMU does not contain modelDescription.xml, which it should according to the FMI specification. """
+        _helper_unzipped_fmu_exception_invalid_dir(FMUModelME1)
+
+    def _test_unzipped_fmu_helper(self, fmu_loader):
```
chria, line 113: Can we rename this? It looks like it tests bouncingBall from an unzipped folder. So I'd rather have _run_bouncing_ball(self, fmi_class) or something similar.

tests/test_fmi.py
```diff
@@ -339,6 +445,14 @@ def test_log_file_name(self):
         path, file_name = os.path.split(full_path)
         assert model.get_log_file_name() == file_name.replace(".","_")[:-4]+"_log.txt"
 
+    @testattr(stddist = True)
+    def test_unzipped_fmu_exceptions(self):
+        """ Verify exception is raised if 'fmu' is a file and allow_unzipped_fmu is set to True, with FMUModelCS2. """
+        err_msg = "Argument named 'fmu' must be a directory if argument 'allow_unzipped_fmu' is set to True."
+        full_path = os.path.join(file_path, "files", "FMUs", "XML", "CS2.0")
+        with nose.tools.assert_raises_regex(FMUException, err_msg):
```
chria, line 453: Also here, why a with?

chria (changes requested):

src/pyfmi/fmi.pyx
modelonrobinandersson: Please review:
https://github.com/modelon-community/PyFMI/commit/09901c64e809fa707c952d070c5381fe24a632e7

src/pyfmi/fmi.pyx
modelonrobinandersson: I agree that the function does not bring much, but I rather see the use of a function instead of the deprecation code duplicated in 5 other places. When we decide later to remove the path argument (as well as the deprecation) I believe it is easier if the developer simply needs to look for the calls to this function. What do you think?

src/pyfmi/fmi.pyx
modelonrobinandersson: Hi, I agree with you, but my reasoning for this is also to reduce code duplication. I guess we have 2 options, using a function or duplicate the code in 5 different locations. I tried to do inline conditionals, i.e. something like:
`f('foo' if conditional else 'bar', ...)` but I didn't get it to work as Cython complained about the type of the temporary variable created within the expression.
Personally, I think we gain more by using this function instead of adding the following expression in 5 different locations:
`if allow_unzipped_fmu:
    f(..., NULL, ...)
else:
    f(..., fmu_full_path, ...)
`
What are your thoughts? In general I think there is a lot of code duplication in pyfmi, and I believe working to fix that would make it easier to work with.

src/pyfmi/fmil_import.pxd
modelonrobinandersson: I referenced the function in fmi.pxd, does it make more sense to have it in fmil_import.pxd?

src/pyfmi/fmi.pyx
modelonrobinandersson: I didn't change the default error message, but perhaps the following would work? Let me know and I'll change it when we have consensus:
"Instantiating an FMU requires an FMU (.fmu) file, specified file has extension {}".format(os.path.splittext(fmu_full_path)[1])"

tests/test_fmi.py
modelonrobinandersson: I used a with-statement since TemporaryDirectory() can be used as a context manager. Basically we don't need to invoke mkdir/rmdir by ourselves.
If you are referring to the exception, it is the preferred way to test exceptions (as far as I know). It will also throw an exception if no exception was thrown.

tests/test_fmi.py
modelonrobinandersson: I agree, it was a poor name for the helper function, changed it here:
https://github.com/modelon-community/PyFMI/commit/33c4c1efaa1b4faa235c2ab25f9b96c91c9ba94d

I also didn't change fmu_loader to fmi_class since for some of the tests we send in the function 'load_fmu' and I was thinking it might be misleading(?)

tests/test_fmi.py
modelonrobinandersson: Same as above, it will also catch scenariso where no exception is thrown with minimal code. I guess the other way would be:
`
exception_raised = False
try:
  foo()
except FMUException as e:
  exception_raised = True
  <verify exception message>
assert exception_raised, "helpful error message"
`
but is a lot more code to achieve the same thing.

src/pyfmi/fmi.pyx
chria: Ok

src/pyfmi/fmi.pyx
chria: What about changing fmu_full_path to a C type and initialize it to NULL? Then if we use it, we set it to the full path, otherwise it is NULL and this call is identical in both cases. I.e.: f(..., fmu_full_path,...)

src/pyfmi/fmil_import.pxd
chria: Everything that reference functions from FMIL should go into fmil_import.pxd

src/pyfmi/fmi.pyx
chria: Looks great!

tests/test_fmi.py
chria: Hmm ok. I was referring to the with statement on "with np.testing.assert_raises_regex" I've not seen this before and don't think we have it anywhere else in the code base. Why is not "np.testing.assert_raises_regex(FMUException, ...., err_msg)" enough?

tests/test_fmi.py
chria: Looks good!

tests/test_fmi.py
chria: Why not just nose.tools.assert_raises(FMUException, function, arguments)?

src/pyfmi/fmi.pyx
modelonrobinandersson: Yes you're right that would make that line identical in all places but then we would have to add logic instead to check if it is used or not and I don't think we gain anything. For example, assume we initialize it to NULL, then I think (correct me if I'm wrong) we need to add code to see if we should set it to os.path.abspath(os.path.join(path,fmu)), that is, if allow_unzipped_fmu is set to True (which adds a new if-block) in all locations where this variable is initialized.

src/pyfmi/fmil_import.pxd
modelonrobinandersson: Oh I see! But the function import_and_get is not a new function from FMIL, it is defined in fmi.pyx, and it uses the same FMIL function as with the old behavior so there is no new FMIL function in this implementation.

tests/test_fmi.py
modelonrobinandersson: As far as I understand, this is the only difference according to their documentation:

> An optional keyword argument 'msg' can be provided when assert_raises
> is used as a context object.

Other than that I just think context managers are better since they allocate and release resources precisely within that block.

tests/test_fmi.py
modelonrobinandersson: See my reply in the other comment regarding the context managers, I think the difference is minor but we also test the error message.

src/pyfmi/fmil_import.pxd
chria: Ok!

tests/test_fmi.py
chria: Hmm ok. Still I must say that I think it becomes harder to read what happens when context managers are introduced.

tests/test_fmi.py
chria: Also here. OK but I think context managers make it harder to read what is being tested.

src/pyfmi/fmi.pyx
chria: Hmm ok, I'd still rather have it as something fmi_import_get_fmi_version(..., NULL if allowed else fmu_path,...) and remove this method.

chria (approved):

PR #91, opened 2021-04-30
Fix numpy 1.20 deprecation warnings

jschueller:
see https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations


jschueller:
@chria  @modelonrobinandersson I think this is pretty much ready to go

PR #94, opened 2021-05-03
Update CHANGELOG

modelonrobinandersson:


CHANGELOG
```diff
@@ -2,6 +2,8 @@
 --- TBD ---
     * Fixed segfault when storing data from models with a huge number of
       variables.
+    * Loading of FMUs can now be done from an unzipped folder if the argument 'allow_unzipped_fmu' is set to True.
+    * The argument 'path' to load_fmu and the different FMI-classes will is now deprecated, a warning will now be displayed when this is set differently from the default value.
```
agnesramle, line 6: I think it is enough with "The argument 'path' to load_fmu and the different FMI-classes is now deprecated."

agnesramle (changes requested):

CHANGELOG
modelonrobinandersson: Fixed, thanks!

PR #95, opened 2021-05-05
Logging into streams

modelonrobinandersson:
This PR covers a new feature that allows for logging into streams as well as tests for it

### DIFF
--- a/src/common/log/parser.py
+++ b/src/common/log/parser.py
@@ -1,7 +1,7 @@
-#!/usr/bin/env python 
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
-# Copyright (C) 2020 Modelon AB
+# Copyright (C) 2020-2021 Modelon AB
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Lesser General Public License as published by
@@ -20,11 +20,12 @@ Parser for a XML based FMU log format
 
 from xml import sax
 import re
+import os
 import numpy as np
 from distutils.util import strtobool
 from .tree import *
 from pyfmi.fmi_util import python3_flag
-
+from pyfmi.fmi import FMUException
 
 ## Leaf parser ##
 
@@ -57,12 +58,12 @@ def parse_value(text):
         if quoted_string_pattern.match(text):
             text = text[1:-1].replace('""','"')
         else:
-            assert '"' not in text            
+            assert '"' not in text
         # for python 2 we need to avoid printing all strings as u'...'
         return text if python3_flag else text.encode('ascii', 'xmlcharrefreplace')
 
 def parse_vector(text):
-    
+
     text = text.strip()
     if text == "":
         return np.zeros(0)
@@ -77,7 +78,7 @@ def parse_matrix(text):
     parts = semicolon_pattern.split(text)
     parts = filter(None, map(str, parts))
     return np.asarray([parse_vector(part) for part in parts])
-    
+
 
 ## SAX based parser ##
 
@@ -113,7 +114,7 @@ class ContentHandler(sax.ContentHandler):
 
     def startElement(self, type, attrs):
         self.create_comment()
-        
+
         key = attrs.get('name')
         # convert to string if key is not None
         # because Python 2.x returns unicode while Python 3.x does not
@@ -121,7 +122,7 @@ class ContentHandler(sax.ContentHandler):
 
         self.chars = []
         self.leafkey = self.leafparser = None
-        
+
         if   type == 'value' or type == 'boolean':  self.leafparser = parse_value
         elif type == 'vector': self.leafparser = parse_vector
         elif type == 'matrix': self.leafparser = parse_matrix
@@ -129,12 +130,12 @@ class ContentHandler(sax.ContentHandler):
         if self.leafparser is not None:
             self.leafkey = key
         else:
-            node = Node(type)            
+            node = Node(type)
             #if len(self.nodes) > 0:
             self.nodes[-1].add(node, key)
             self.nodes.append(node)
-            
-    def endElement(self, type):        
+
+    def endElement(self, type):
         # todo: verify name matching?
         if self.leafparser is not None:
             node = self.leafparser(self.take_chars())
@@ -168,7 +169,7 @@ def parse_xml_log(filename, accept_errors=False):
             print('Parsed log will be incomplete.')
         else:
             raise Exception('Failed to parse XML FMU log:\n' + repr(e))
-        
+
     return handler.get_root()
 
 
@@ -192,19 +193,42 @@ def parse_fmu_xml_log(filename, modulename = 'Model', accept_errors=False):
             print('Parsed log will be incomplete')
         else:
             raise Exception('Failed to parse XML FMU log:\n' + repr(e))
-    
+
     return handler.get_root()
 
-def extract_xml_log(destfilename, filename, modulename = 'Model'):
+def extract_xml_log(destfilename, log, modulename = 'Model'):
     """
     Extract the XML contents of a FMU log and write as a new file destfilename.
-
-    modulename selects the module as recorded in the beginning of each line by
+    Input argument 'modulename' selects the module as recorded in the beginning of each line by
     FMI Library.
+
+        destfilename::
+
+            file_name --
+                Name of the file which holds the extracted log
+                Default: get_log_filename() + xml
+            log --
+                String of filename to extract log from or a stream. The support for stream
+                requires that the stream is readable and supports the attributes 'seek' and 'readlines'.
+                For information about these attributes, see the class
+                IOBase in the module 'io', that is part of the Python standard library.
+                Default: get_log_filename() + xml
+            modulename --
+                Selects the module as recorded in the beginning of each line by FMI Library.
+                Default: 'Model'
     """
-    with open(filename, 'r') as sourcefile:
-        with open(destfilename, 'w') as destfile:
-            filter_xml_log(destfile.write, sourcefile, modulename)
+    if isinstance(log, str):
+        with open(log, 'r') as sourcefile:
+            with open(destfilename, 'w') as destfile:
+                filter_xml_log(destfile.write, sourcefile, modulename)
+    elif hasattr(log, 'seek') and hasattr(log, 'readlines'):
+            log.seek(0) # Return to the start of the stream
+            with open(destfilename, 'w') as destfile:
+                filter_xml_log(destfile.write, log.readlines(), modulename)
+    else:
+        msg = "Input argument 'log' needs to be either a file or a stream that supports"
+        msg += " the two attributes 'seek' and 'readlines'."
+        raise FMUException(msg)
 
 def filter_xml_log(write, sourcefile, modulename = 'Model'):
     write('<?xml version="1.0" encoding="UTF-8"?>\n<JMILog category="info">\n')
--- a/src/common/log/parser.py
+++ b/src/common/log/parser.py
@@ -202,9 +202,9 @@ def extract_xml_log(dest, log, modulename = 'Model'):
     Input argument 'modulename' selects the module as recorded in the beginning of each line by
     FMI Library.
 
-        dest::
+        parameters::
 
-            file_name  --
+            dest  --
                 Name of the file which holds the extracted log, or a stream to write to
                 that supports the function 'write'. Default behaviour is to write to a file.
                 Default: get_log_filename() + xml
--- a/src/common/log/parser.py
+++ b/src/common/log/parser.py
@@ -229,17 +229,16 @@ def extract_xml_log(dest, log, modulename = 'Model'):
                     filter_xml_log(destfile.write, sourcefile, modulename)
             else:
                 filter_xml_log(dest.write, sourcefile, modulename)
-    elif hasattr(log, 'seek') and hasattr(log, 'readlines'):
-            log.seek(0) # Return to the start of the stream
+    elif hasattr(log, 'readlines'):
             if dest_is_file:
                 with open(dest, 'w') as destfile:
                     filter_xml_log(destfile.write, log.readlines(), modulename)
             else:
                 filter_xml_log(dest.write, log.readlines(), modulename)
     else:
-        msg = "Input argument 'log' needs to be either a file or a stream that supports"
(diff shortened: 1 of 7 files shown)


src/common/log/parser.py
```diff
@@ -192,19 +193,53 @@ def parse_fmu_xml_log(filename, modulename = 'Model', accept_errors=False):
...
-    Extract the XML contents of a FMU log and write as a new file destfilename.
-
-    modulename selects the module as recorded in the beginning of each line by
+    Extract the XML contents of a FMU log and write as a new file dest, or extract into a stream.
+    Input argument 'modulename' selects the module as recorded in the beginning of each line by
     FMI Library.
+
+        dest::
+
+            file_name  --
```
chria, line 207: I think file_name should be removed right? It should say only dest?

src/common/log/parser.py
```diff
@@ -192,19 +193,53 @@ def parse_fmu_xml_log(filename, modulename = 'Model', accept_errors=False):
...
+            else:
+                filter_xml_log(dest.write, sourcefile, modulename)
+    elif hasattr(log, 'seek') and hasattr(log, 'readlines'):
+            log.seek(0) # Return to the start of the stream
+            if dest_is_file:
+                with open(dest, 'w') as destfile:
+                    filter_xml_log(destfile.write, log.readlines(), modulename)
+            else:
+                filter_xml_log(dest.write, log.readlines(), modulename)
+    else:
```
chria, line 239: Why? I think we should try to assume that the stream is at the start. We should not assume to need to seek in the stream.

src/pyfmi/fmi.pxd
```diff
@@ -41,9 +41,12 @@ cdef class ModelBase:
     cdef char* _fmu_log_name
     cdef FMIL.jm_callbacks callbacks
     cdef public dict cache
+    cdef public object _log_stream
     cdef public object file_object
     cdef public object _additional_logger
     cdef public object _max_log_size_msg_sent
+    cdef object _modelid
+    cdef public int log_is_stream, _invoked_dealloc
```
chria, line 49: I suggest to make "log_is_stream" to an internal variable i.e: "_log_is_stream"

src/pyfmi/fmi.pyx
```diff
@@ -469,15 +482,19 @@ cdef class ModelBase:
         return self._default_options('pyfmi.fmi_algorithm_drivers', algorithm)
 
     def get_log_filename(self):
-        return decode(self._fmu_log_name)
+        """ Returns a name of the logfile, if logging is done to a stream it returns
+            a string formatted base on the model identifier.
+        """
+        return '{}_log'.format(self._modelid) if self.log_is_stream else decode(self._fmu_log_name)
```
chria, line 488: Why do we return a filename for when it is a stream? There is no file... (or at least any file that we know / or care about)

src/pyfmi/fmi.pyx
```diff
@@ -496,33 +513,49 @@ cdef class ModelBase:
         for i in range(N):
             print(log[i])
 
-    def extract_xml_log(self, file_name=None):
+    def extract_xml_log(self, file_name=None, stream = None):
```
chria, line 516: Why do we need both? Should we not just have the functionality to extract the log to a user-defined place, be it either a file or a stream. So can we not just have one argument?

src/pyfmi/fmi.pyx
```diff
@@ -496,33 +513,49 @@ cdef class ModelBase:
...
         Extract the XML contents of a FMU log and write as a new file.
 
         Parameters::
 
             file_name --
-                Name of the file which holds the extracted log
+                Name of the file which holds the extracted log, or a stream to write to
+                that supports the function 'write'. Default behaviour is to write to a file.
                 Default: get_log_filename() + xml
+            stream --
```
chria, line 526: Hmm I see, what is the use-case here? We already have the stream where we've stored the data?

src/pyfmi/fmi.pyx
```diff
@@ -560,16 +596,44 @@ cdef class ModelBase:
...
+        elif self._log_stream_is_open():
+            return True
         else:
             return False
 
+    def _log_stream_is_open(self):
+        """ Returns True or False based on if logging is done to a stream and if it is open or closed. """
+        return self._log_stream_is_set() and not self._log_stream.closed
+
+    def _log_stream_is_set(self):
```
chria, line 630: Lets remove this method.

src/pyfmi/fmi.pyx
```diff
@@ -607,21 +671,35 @@ cdef class ModelBase:
                     if (f != NULL):
                         FMIL.fprintf(f, "FMIL: module = %s, log level = %d: %s\n",c_module, log_level, c_message)
                         FMIL.fclose(f)
+        elif self._log_stream:
+            try:
+                self._log_stream.write(msg)
+            except:
```
chria, line 677: Maybe a clarifying comment about what we are capturing here?

src/pyfmi/fmi.pyx
```diff
@@ -1359,15 +1440,27 @@ cdef class FMUModelBase(ModelBase):
...
+            self._set_log_stream(log_file_name)
             for i in range(len(self._log)):
-                file.write("FMIL: module = %s, log level = %d: %s\n"%(self._log[i][0], self._log[i][1], self._log[i][2]))
-            self._log = []
+                try:
+                    self._log_stream.write("FMIL: module = %s, log level = %d: %s\n"%(self._log[i][0], self._log[i][1], self._log[i][2]))
+                except:
+                    if hasattr(self._log_stream, 'closed') and self._log_stream.closed:
+                        logging.warning("Unable to log to closed stream.")
+                    logging.warning("Unable to log to stream.")
```
chria, line 1452: Should this not be in an else clause? Otherwise we get two warnings if the log has a closed attribute

src/pyfmi/fmi.pyx
```diff
@@ -2704,6 +2797,10 @@ cdef class FMUModelCS1(FMUModelBase):
         """
         Deallocate memory allocated
         """
+        self._invoked_dealloc = 1
+        if self._log_stream:
```
chria, line 2801: Hmm, why is this done here? It seems like if you have a stream, you remove it directly?

src/pyfmi/fmi.pyx
```diff
@@ -3266,6 +3363,10 @@ cdef class FMUModelME1(FMUModelBase):
...
```
chria, line 3367: Same here, this will make it impossible to log to the stream in the terminate / free methods.

src/pyfmi/fmi.pyx
```diff
@@ -4004,25 +4108,32 @@ cdef class FMUModelBase2(ModelBase):
 
         #Load information from model
         if isinstance(self,FMUModelME2):
-            self._modelId           = decode(FMIL.fmi2_import_get_model_identifier_ME(self._fmu))
+            self._modelid           = decode(FMIL.fmi2_import_get_model_identifier_ME(self._fmu))
```
chria, line 4111: Why change to lower case id? We usually use a capital letter to indicate a new name/word. I.e. here we have "model" and "id" -> modelId.

src/pyfmi/fmi.pyx
```diff
@@ -30,6 +30,7 @@ import fnmatch
 import re
 from collections import OrderedDict
 cimport cython
+from io import StringIO, TextIOWrapper, UnsupportedOperation
```
chria, line 33: Doesn't seem like we need these?

tests/test_stream.py
```diff
@@ -0,0 +1,280 @@
+#!/usr/bin/env python
```
chria, line 1: What about tests for the async io?

src/common/log/parser.py
modelonrobinandersson: Correct this is just an error due to search/replace. I've now fixed this.

src/common/log/parser.py
modelonrobinandersson: I would be OK with that but that would also mean that the following workflow is not going to work:
```
fmu.simulate()
fmu.extract_xml_log()
```
Because the user would have to reset the stream postion in-between, wouldn't it be valuable to keep this? Or what do you think?

src/pyfmi/fmi.pyx
modelonrobinandersson: I initially changed this function to return an empty string, but then when I looked at extract_xml_log, it says that the file it creates is based on what this function is returning and therefore I changed it to above.
I could however change this function to return an empty string instead and update to docstring in extract_xml_log, what approach do you think is the best?

src/pyfmi/fmi.pxd
modelonrobinandersson: I agree, it has been fixed.

src/pyfmi/fmi.pyx
modelonrobinandersson: I added the two arguments because the user can with argument 'stream' now extract_xml_log either from a separate stream or from the the internal stream from the loading of the fmu, i.e.:
```
stream1 = ...
stream2 = ... # some empty stream
stream3 = ...
fmu = load_fmu(compiled_fmu, log_filename = stream1)
fmu.simulate(a, b)
fmu = load_fmu(compiled_fmu, log_filename = stream3)
fmu.simulate(a, c)
fmu.extract_xml_log() # will extract from stream3 and write to a file as default
fmu.extract_xml_log(file_name = stream2) # will extract from stream3 but extract it into stream2 instead of generating a file.
fmu.extract_xml_log(stream = stream1) # will extract from stream1 and generate a file
```

src/pyfmi/fmi.pyx
modelonrobinandersson: Yes correct, see my comment above where I wrote an example with 3 streams, if it doesn't make sense I can change it but that is the reason why I added it.

src/pyfmi/fmi.pyx
modelonrobinandersson: Fixed

src/pyfmi/fmi.pyx
modelonrobinandersson: Fixed

src/pyfmi/fmi.pyx
modelonrobinandersson: Good catch, fixed as well

src/pyfmi/fmi.pyx
modelonrobinandersson: I thought this is what we agreed to during a separate discussion but I probably misunderstood this because I agree with the second comment also about the terminate/free methods. I've updated it to do this last instead (i.e. after we free self._fmu_log_name).

src/pyfmi/fmi.pyx
modelonrobinandersson: Fixed as well (see above)

src/pyfmi/fmi.pyx
modelonrobinandersson: I agree, an error on my side, I've fixed it

src/pyfmi/fmi.pyx
modelonrobinandersson: Correct, I've removed the first 2 (we still need the UnsupportedOperation)

tests/test_stream.py
modelonrobinandersson: I didn't add an explicit test with asyncio but I included it as a part of tests with the class TestIO, since if we a user were to give a stream from asyncio they would need to inherit in their own class in order to invoke drain after each write. That is, the flow with asyncio-stream is
```
stream.write(...)
stream.drain()
``` 
But in PyFMI we only invoke write, so they'd have to have a class inheriting their own write function, that is why I added the testclass TestIO but it instead inherits from StringIO due to simplicity.

src/pyfmi/fmi.pyx
chria: I see, if the user wants to extract the xml from a separate stream, they should use directly the "extract_xml_log" from the common package. I don't think we should support it here. I.e. the last case should not be supported.

src/pyfmi/fmi.pyx
chria: Hmm, does not seem that the else is added?

tests/test_stream.py
chria: I see, do we have a comment that this is needed somewhere?

src/common/log/parser.py
modelonrobinandersson: I agree that makes sense:
See https://github.com/modelon-community/PyFMI/commit/8e355093157fb6e9b1824156bc923aa8a873cb86

src/pyfmi/fmi.pyx
modelonrobinandersson: Fixed with the comment above (where I linked to this changeset https://github.com/modelon-community/PyFMI/commit/8e355093157fb6e9b1824156bc923aa8a873cb86)

src/pyfmi/fmi.pyx
modelonrobinandersson: Ah sorry there were two places in the code with this exact code, added in the second place as well:
https://github.com/modelon-community/PyFMI/commit/5103e9188ad872431f37e08004298ccde48f17ad

tests/test_stream.py
modelonrobinandersson: We don't. We've only stated that the streams need to support write, do you think it makes sense if we do the following change?
CURRENT (docstring for load_fmu and the FMI-classes):
```
        log_file_name --
            Filename for file used to save logmessages.
            This argument can also be a stream if it supports 'write', for full functionality
            it must also support 'seek' and 'readlines'.
            The stream must also be open and writable during the entire time.
            Default: "" (Generates automatically)
```
PROPOSAL:
```
        log_file_name --
            Filename for file used to save logmessages.
            This argument can also be a stream if it supports 'write', for full functionality
            it must also support 'seek' and 'readlines'. If the stream requires use of other methods, such as 'drain' 
            for asyncio-streams, then this needs to be implemented on the user-side, there is no additional methods invoked 
            on the stream instance after 'write' has been invoked on the PyFMI side.
            The stream must also be open and writable during the entire time.
            Default: "" (Generates automatically)
```

chria: Looks good!

modelonrobinandersson: Thanks! Added now with the latest commit.

PR #101, opened 2021-05-18
Support for writing results to streams and also bug fix for on demand loading with 32 bit data in result

agnesramle:


src/common/io.py
```diff
@@ -231,172 +231,192 @@ def _find_underlying_state(self, name):
...
+                Default: ";""
+        """
+
+        if isinstance(filename, str):
+            fid = codecs.open(filename,'r','utf-8')
+        else: #assume stream
+            if not(hasattr(filename, 'readline') and hasattr(filename, 'seek')):
+                raise JIOError("Given stream needs to support 'readline' and 'seek' in order to retrieve the results.")
+            fid = filename
+            fid.seek(0, 0) # need to start reading from beginning
```
chria, line 290: Can we assume that we start from the beginning or do we need this?

src/common/io.py
```diff
@@ -1547,11 +1580,18 @@ def simulation_start(self):
...
         # Open file
-        f = codecs.open(self.file_name,'w','utf-8')
-        self.file_open = True
-        
+        if isinstance(self.file_name, str):
+            f = codecs.open(self.file_name,'w','utf-8')
+            self.file_open = True
+        else:
+            if not hasattr(self.file_name, 'write'):
+                raise fmi.FMUException("Instance set via option 'result_file_name' needs to support 'write' as required.")
```
chria, line 1590: Not sure that I understand the exception here. What about something like "Failed to write the result file. The provided "result_file_name" needs to be a filename or a class that supports writing to through the 'write' method."

src/common/io.py
```diff
@@ -1657,55 +1697,62 @@ def set_options(self, options):
...
         parameters = self.parameters
-        
+
         # Open file
-        f = codecs.open(file_name,'w','utf-8')
+        if isinstance(self.file_name, str):
+            f = codecs.open(self.file_name,'w','utf-8')
+        else:
+            if not (hasattr(self.file_name, 'write') and hasattr(self.file_name, 'seek')):
+                raise fmi.FMUException("Instance set via option 'result_file_name' needs to support 'write' and 'seek' as required.")
```
chria, line 1744: Same regarding the error message as above.

src/common/io.py
```diff
@@ -2182,160 +2234,169 @@ def _data_header(self, name, nbr_rows, nbr_cols, data_type):
...
+
         file_name = self.file_name
         parameters = self.parameters
-        self._file = open(file_name,'wb')
-        
+        if isinstance(self.file_name, str):
+            self._file = open(file_name,'wb')
+        else:
+            if not (hasattr(self.file_name, 'write') and hasattr(self.file_name, 'seek') and (hasattr(self.file_name, 'tell'))):
+                raise fmi.FMUException("Instance set via option 'result_file_name' needs to support 'write', 'tell' and 'seek' as required.")
```
chria, line 2311: Same as above regarding the error message.

src/pyfmi/fmi_util.pyx
```diff
@@ -1089,12 +1088,58 @@ def read_trajectory(file_name, int data_index, int file_position, int sizeof_typ
     cdef unsigned long int start_point = data_index*sizeof_type
     cdef unsigned long int end_point   = sizeof_type*(nbr_points*nbr_variables)
     cdef unsigned long int interval    = sizeof_type*nbr_variables
+
+    if sizeof_type == 4:
+        return _read_trajectory32(file_name, start_point,end_point, interval, file_position, nbr_points)
+    elif sizeof_type == 8:
+        return _read_trajectory64(file_name, start_point,end_point, interval, file_position, nbr_points)
+    else:
+        raise JIOError("Cannot read data that is not of size 32 or 64 bit (float).")
```
chria, line 1097: What about adding something more to the message: "Failed to read the result. The result is on an unsupported format. Can only read data that is either a 32 or 64bit double."

src/pyfmi/fmi_util.pyx
```diff
@@ -1089,12 +1088,58 @@ def read_trajectory(file_name, int data_index, int file_position, int sizeof_typ
...
+
+DTYPE32 = np.float32
+ctypedef np.float32_t DTYPE32_t
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def _read_trajectory32(file_name, long int start_point, long int end_point, long int interval, int file_position, int nbr_points):
```
chria, line 1104: Propose to define this as a "cdef" method (for performance) as it is only called from this file

src/pyfmi/fmi_util.pyx
```diff
@@ -1089,12 +1088,58 @@ def read_trajectory(file_name, int data_index, int file_position, int sizeof_typ
...
+    fclose(cfile)
+
+    return data
+
+DTYPE = np.double
+ctypedef np.double_t DTYPE_t
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+def _read_trajectory64(file_name, long int start_point, long int end_point, long int interval, int file_position, int nbr_points):
```
chria, line 1133: Same as above, propose to change this to a cdef.

chria (commented):
Good job!

src/common/io.py
agnesramle: It depends on what kind of user experience one wants. If we aren't at the beginning of the file we may get unexpected results back which will be hard for the user to trace.

src/pyfmi/fmi_util.pyx
agnesramle: Fixed.

src/pyfmi/fmi_util.pyx
agnesramle: Fixed.

src/common/io.py
agnesramle: Fixed

src/common/io.py
agnesramle: Fixed

src/common/io.py
chria: Ok

chria:
Looks good! (Conflict needs to be resolved though)

PR #102, opened 2021-05-24
Updated exception types

chria:


src/pyfmi/fmi.pyx
```diff
@@ -4087,26 +4110,26 @@ cdef class FMUModelBase2(ModelBase):
...
-                raise FMUException("The FMU kind could not be determined. Enable logging for possibly more information.")
+                raise InvalidVersionException("The FMU kind could not be determined. Enable logging for possibly more information.")
         elif self._fmu_kind == FMIL.fmi2_fmu_kind_me_and_cs:
             if isinstance(self,FMUModelME2):
                 self._fmu_kind = FMIL.fmi2_fmu_kind_me
             elif isinstance(self,FMUModelCS2):
                 self._fmu_kind = FMIL.fmi2_fmu_kind_cs
             else:
-                raise FMUException("FMUModelBase2 cannot be used directly, use FMUModelME2 or FMUModelCS2.")
+                raise InvalidVersionException("FMUModelBase2 cannot be used directly, use FMUModelME2 or FMUModelCS2.")
```
agnesramle, line 4122: I'm not convinced this should be an InvalidVersionException, version is 2 in all three cases here.

src/pyfmi/fmi.pyx
```diff
@@ -4115,7 +4138,7 @@ cdef class FMUModelBase2(ModelBase):
         elif isinstance(self,FMUModelCS2):
             self._modelId           = decode(FMIL.fmi2_import_get_model_identifier_CS(self._fmu))
         else:
-            raise FMUException("FMUModelBase2 cannot be used directly, use FMUModelME2 or FMUModelCS2.")
+            raise InvalidFMUException("FMUModelBase2 cannot be used directly, use FMUModelME2 or FMUModelCS2.")
```
agnesramle, line 4141: Same as above.

src/pyfmi/fmi.pyx
```diff
@@ -6969,10 +6992,10 @@ cdef class FMUModelCS2(FMUModelBase2):
 
         if self._fmu_kind != FMIL.fmi2_fmu_kind_cs:
             if self._fmu_kind != FMIL.fmi2_fmu_kind_me_and_cs:
-                raise FMUException("This class only supports FMI 2.0 for Co-simulation.")
+                raise InvalidVersionException("This class only supports FMI 2.0 for Co-simulation.")
 
         if self.get_capability_flags()['needsExecutionTool'] == True:
-            raise FMUException('Models that need an execution tool are not supported')
+            raise InvalidVersionException('Models that need an execution tool are not supported')
```
agnesramle, line 6998: Why should this be an InvalidVersionException? needsExecutionTool has nothing to do with version.

src/pyfmi/fmi.pyx
```diff
@@ -7598,10 +7621,10 @@ cdef class FMUModelME2(FMUModelBase2):
 
         if self._fmu_kind != FMIL.fmi2_fmu_kind_me:
             if self._fmu_kind != FMIL.fmi2_fmu_kind_me_and_cs:
-                raise FMUException('This class only supports FMI 2.0 for Model Exchange.')
+                raise InvalidVersionException('This class only supports FMI 2.0 for Model Exchange.')
 
         if self.get_capability_flags()['needsExecutionTool'] == True:
-            raise FMUException('Models that need an execution tool are not supported')
+            raise InvalidVersionException('Models that need an execution tool are not supported')
```
agnesramle, line 7627: Same as above.

agnesramle (changes requested):
I miss updates of the tests to catch exceptions of the 'child' type and not the 'parent' type. I see it as important since that will also make sure that the exception types are possible to import.
I think there is at least one update in master.pyx needed as well:
raise fmi.FMUException("The Master algorithm currently only supports CS 2.0 FMUs.")
should be InvalidVersionException I think.
The same goes for
raise fmi.FMUException("The coupled model currently only supports ME 2.0 FMUs.")
in fmi_coupled.pyx

src/pyfmi/fmi.pyx
chria: I agree, it is not ideal. Any suggestion?

agnesramle: For now, I suggest to keep the FMUException.

chria:
Updated after review. We don't have that many tests that catch specific types (and not that many regarding loading exceptions). Updated the one I found.

agnesramle:
That we don't have tests regarding loading exceptions is bad. I suggest to add at least one test per exception type here. That will at least test that it is possible to import and catch these Exceptions.

chria:
Updated and added a few tests.

agnesramle:
LGTM!

PR #103, opened 2021-05-25
Resolved issue with logfile being created even if no logging was done

modelonrobinandersson:


src/pyfmi/fmi.pyx
```diff
@@ -625,14 +631,15 @@ cdef class ModelBase:
...
+        """ Sets parameter to keep log file open when logging is done to log file, if we are not logging into a given stream. """
         if not self._log_is_stream and self._fmu_log_name != NULL:
-            self.file_object = open(self._fmu_log_name,'a')
+            self._keep_log_open = 1
 
     def _close_log_file(self):
         if self.file_object:
             self.file_object.close()
             self.file_object = None
+            self._keep_log_open = 0
```
chria, line 642: I propse to always set this in this function. I.e. move self._keep_log_open = 0 outside of the if-statement

tests/test_log.py
```diff
@@ -100,3 +105,62 @@ def test_extract_boolean_value(self):
         eis = log.find("EventInfo")
         for ei in eis:
             assert isinstance(ei.time_event_info, bool), "Expected ei.time_event_info to be bool"
+
+class TestNoLogFile:
```
chria, line 109: Does these tests really work?

chria (commented):
Looks good! Just a few comments

src/pyfmi/fmi.pyx
modelonrobinandersson: Good point, I've updated the code with the latest commit

tests/test_log.py
modelonrobinandersson: Yes they do, they run without issues when I run them locally, or is it anything you see that indicates that they should not work?

tests/test_log.py
```diff
@@ -100,3 +105,62 @@ def test_extract_boolean_value(self):
...
+    """
+        Test invoking functions on the FMU if we have no logfile.
+        We do this from temporary directories to make sure that we have empty directories for the test.
+    """
+    @classmethod
+    def setup_class(cls):
+        cls.fmu_path = os.path.join(get_examples_folder(), 'files', 'FMUs', 'ME2.0', 'bouncingBall.fmu')
```
chria, line 116: Should this not be: cls.fmu_path = os.path.join(get_examples_folder(), 'files', 'FMUs', 'XML', 'ME2.0', 'bouncingBall.fmu') ?

tests/test_log.py
chria: Ok, see comment below.

tests/test_log.py
modelonrobinandersson: That would be the case if it was the folder 'files' in 'tests' but now this FMU is picked from the example folder instead and the path is slightly different there.

chria:
Looks good!

PR #105, opened 2021-05-26
Revert "Resolved issue with logfile being created even if no logging was done"

modelonrobinandersson:
Reverts modelon-community/PyFMI#103 because of issues detected with log files when FMU is loaded and simulated in a loop. If you load + simulate followed by load and simulate, the logging from the second simulation is appended into the same logfile as the first simulation.


modelonrobinandersson:
Reverted due to issues that need to be looked into further before we can add this

Issue #72, opened 2021-01-25
Event handling for ME2 on consecutive simulations with different inputs


dhblum:
I wanted to report an issue with simulating and ME2 FMU in consecutive steps without initialization between steps where the input changes from one step to the next.  For example, initialize and simulate from 0 to 100 s with input u1 = 0, then simulate (without initialization) from 100 to 200 s with input u1 = 1.  There seems to be a problem with identifying and accounting for events that may be triggered by this change in input from one step to the next.  

I think I've identified a problem at line 345 of ``fmi_algorithms_drivers.py`` ([here](https://github.com/modelon-community/PyFMI/blob/master/src/pyfmi/fmi_algorithm_drivers.py#L345)).  Note that 346 performs an event check if the FMU is ME1, however, does not do this check if the FMU is ME2, thereby seemingly skipping this check altogether.  My proposal is to insert the following code after line 349:

```
        elif isinstance(self.model, fmi.FMUModelME2):
            self.model.event_update()    
```

I created and attached to this comment a test Modelica model that can be compiled into an FMU and then simulated using the python script.  The Modelica model looks like the following:  

![image](https://user-images.githubusercontent.com/19480999/105731232-3262ba80-5efd-11eb-9be1-b9477d275d4d.png)

The results of running the Python script with and without the fix I propose looks like the following:

![image](https://user-images.githubusercontent.com/19480999/105731320-4b6b6b80-5efd-11eb-9502-7139d1988e75.png)

I am not sure if I am missing something that either makes this a non-issue or if there's more to be done to fix, but I wanted to report and see what the developers and community thought.

I tested using Python 2.7.17, pyfmi v2.5, on Ubuntu 18.04.5 LTS.

[eventTest_mo.txt](https://github.com/modelon-community/PyFMI/files/5867819/eventTest_mo.txt)
[eventTest_py.txt](https://github.com/modelon-community/PyFMI/files/5867820/eventTest_py.txt)

jschueller:
maybe you'll want to submit a PR ?

dhblum:
Thanks @jschueller, done in https://github.com/modelon-community/PyFMI/pull/109.

modelonrobinandersson:
@dhblum I will close this issue since you have opened the PR for this, any future discussions can be handled in the PR.

PR #109, opened 2021-06-09
Add event check and update for ME2

dhblum:
This is for #72.  

In particular, it invokes an event update for an ME2 model at the start time of a simulation.  Currently, it looks like this is only done for ME1.

### DIFF
--- a/src/pyfmi/fmi_algorithm_drivers.py
+++ b/src/pyfmi/fmi_algorithm_drivers.py
@@ -351,7 +351,9 @@ class AssimuloFMIAlg(AlgorithmBase):
             event_info = self.model.get_event_info()
             if event_info.upcomingTimeEvent and event_info.nextEventTime == model.time:
                 self.model.event_update()
-        
+        elif isinstance(self.model, fmi.FMUModelME2):
+            self.model.event_update()
+
         if abs(start_time - model.time) > 1e-14:
             logging.warning('The simulation start time (%f) and the current time in the model (%f) is different. Is the simulation start time correctly set?'%(start_time, model.time))
         


src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -351,7 +351,9 @@ def __init__(self,
             event_info = self.model.get_event_info()
             if event_info.upcomingTimeEvent and event_info.nextEventTime == model.time:
                 self.model.event_update()
-        
+        elif isinstance(self.model, fmi.FMUModelME2):
+            self.model.event_update()
```
jschueller, line 355: what about the event_info condition ?
I see fmi2 also has event_info but upcomingTimeEvent is named nextEventTimeDefined

dhblum: Thanks @jschueller!  And really sorry for my delay here.  But after some further testing compiling my test fmu (from https://github.com/modelon-community/PyFMI/issues/72) as an ME1 instead of ME2, it seems the ``if`` condition in Line 352 is also insufficient to detect an event.  I'm finding that ``event_info.upcomingTimeEvent`` always returns false and ``event_info.nextEventTime`` always returns 0.0.  Similarly, if applied to ME2, ``event_info.nextEventTimeDefined`` always returns false and ``event_info.nextEventTime`` always returns 0.0.  Therefore, in order to get my expected performance, the ``self.model.event_update()`` needs to be called as I suggest, without the ``if``condition on ``event_info``, in both the ME1 and ME2 cases.  I'm still not sure if I'm missing something here though.

modelonrobinandersson:
@dhblum Hi, I am sorry for the lack of replies. Do you know if this is still an issue? I am trying to look into the current issues and close obsolete ones and mark still active ones.

dhblum:
Thanks @modelonrobinandersson.  As far as I can tell, yes this is still an issue, as in the master branch the section starting at https://github.com/modelon-community/PyFMI/blob/master/src/pyfmi/fmi_algorithm_drivers.py#L390 still only checks if there's an event if FMU is ME1 but not if ME2.

modelonrobinandersson:
Thank you @dhblum for the update. I will have to look into the FMI spec first and figure out if this is intended or a miss. I will get back to you as soon as I know more but it will likely be later in August.

dhblum:
Thank you @modelonrobinandersson for your attention to this, sounds good.

modelonrobinandersson:
Hey @dhblum, sorry for getting back to this a bit late. Do you have a test model in mind where this change seems to be necessary?

jschueller:
@modelonrobinandersson the test case can be found attached in #72 if I'm not mistaken

modelonrobinandersson:
> @modelonrobinandersson the test case can be found attached in #72 if I'm not mistaken

Thank you, I will take a look.

dhblum:
Thank you @modelonrobinandersson for taking a look.

dhblum:
Alright, thanks so much @modelonrobinandersson.

modelonrobinandersson:
Hi @dhblum! I have now looked into it further. The good news is that it is working as expected if you try to simulate the full time directly (instead of advancing step by step). See below:
```
    u1 = [0, 0, 0.7, 0.7, 0.1]
    u2 = [0, 0, 2, 2, 0.5]
    u3 = [0, 0, 3.3, 3.3, 2.3]
    inputs = (u1, u2, u3)
    time = np.array([0, 100, 100, 200, 200])
    input_object = (['u1','u2', 'u3'], np.transpose(np.vstack((time, inputs[0], inputs[1], inputs[2]))))
    fmu_path = compile_fmu('eventTest', 'eventTest.mo', target='me')
    fmu = load_fmu(fmu_path)
    res = fmu.simulate(0, 300, input = input_object) 
```
The reason why it does not work the way you have done in your script is because there is currently an issue where we need to detect if there are state events when the simulation starts. This is not trivial to fix. I don't have any time for when we can look into this again except for some time again in the future.

modelonrobinandersson:
> doesnt this PR actually fix the issue ?

Hi. Unfortunately it does not because we only want to do the event update if there are state events, the PR however forces event updates every time the simulation starts which is incorrect. Moreover two of the PyFMI tests that are working as expected fail with the the PR.

modelonrobinandersson:
Hi @lambtt, sorry for my late reply I just discovered I have not replied. We will fix this issue at some point but currently we haven't been able to prioritize it. It also seems that this issue is quite rare as it is not commonly appearing in most FMUs.
I also saw your comment in https://github.com/modelon-community/PyFMI/issues/159#issuecomment-1345820886 and I will therefore close issue #159 as I have now responded here.

Again, sorry for the very late response.

wehuang16:
Hey everyone, I have been using pyfmi for a while and arrived at the same Functional Mockup Unit (FMU) issue on this pull request. Several people on my team and I are working on “model predictive controls” with FMUs, which requires us to take the outputs of an FMU at the current time step, feed them into a “model predictive control” algorithm, and return the desired inputs into the FMU at the next time step (while maintaining the states of the FMU at every time step). Therefore, being able to advance step by step with the pyfmi/FMU simulation while maintaining the FMU states would be  very helpful for our “model predictive control”  algorithm development, as compared to simulating the FMU for the full time. 

I appreciate the efforts of everyone to identify this issue, and hope this bug can be prioritized soon:)

jschueller:
@modelonrobinandersson I propose a more suitable fix in #412, could you review ?

PR #114, opened 2021-07-28
Saving diagnostic data in binary result file instead

agnesramle:


src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -543,6 +552,73 @@ def _set_solver_options(self):
...
+                ei_values = self.model.get_event_indicators() if nof_ei > 0 else []
+                for i in range(nof_ei):
+                    self._diagnostics_var.append((diagnostics_start_name+"event_info.indicator_"+str(i+1), "Value for event indicator {}.".format(i+1)))
+                    self._diagnostics_var_start_values.append(ei_values[i])
+                for i in range(nof_ei):
+                    self._diagnostics_var.append((diagnostics_start_name+"event_info.state_event_info.index_"+str(i+1), "Zero crossing indicator for event indicator {}".format(i+1)))
+                    self._diagnostics_var_start_values.append(0.0)
+                self._diagnostics_var.append((diagnostics_start_name+"event_info.event_type", "No event=-1, state event=0, time event=1")) 
+                self._diagnostics_var_start_values.append(-1) 
+
```
jespermattsson, line 621: Perhaps you should have the diagnostics values as a map name → value or similar instead of a list of names and a list of values? The current approach seems both more fragile and less logical. If you later need it as a list, it is easy enough to convert. I see that the “names” one is actually [(name, description)] - perhaps you could do {name: (description, value)} or [(name, description, value)]?
There are a lot of places where you pass on all three diagnostics_* arguments on to other functions. It gets very verbose and in the way of readability. You may want to consider an object holding all three of them - then you could just pass that around.
You could also combine both of these to have an object with a map (or whatever is used for diagnostics_param_names + diagnostic_param_values) and a list for diagnostics_var.

src/pyfmi/fmi_util.pyx
```diff
@@ -310,7 +313,22 @@ cpdef prepare_data_info(np.ndarray[int, ndim=2] data_info, list sorted_vars, mod
...
+
+    last_index = 0    
+    for i in range(nof_sorted_vars+1+nof_diag_params, nof_sorted_vars+1+nof_diag_params+nof_diag_vars):
+        data_info[0,i] = 3
+        data_info[2,i] = 0
+        data_info[3,i] = -1
+        last_index = last_index + 1
+        data_info[1,i] = last_index
+
+    data = np.append(np.append(np.append(np.append(model.time, model.get_real(param_real)),model.get_integer(param_int).astype(float)),model.get_boolean(param_bool).astype(float)), np.array(diagnostics_param_values).astype(float))
```
jespermattsson, line 331: Sorry, but why on earth is that many calls nestled like that?
I see that there was a chain of append calls before and you just added one more, but you should still clean up things like this when you need to change them.
Looking at the docs ([numpy.append from NumPy 1.21](https://numpy.org/doc/stable/reference/generated/numpy.append.html), I hope that is the correct one), append accepts a list for the second argument. As far as I can tell, just calling it with all except the first argument lumped into a list will give the same results. There is also an axis arguments that lets you add columns or rows to a matrix and similar. 
Do note that each nestled call to append creates a new copy of the entire array.

src/pyfmi/fmi_util.pyx
```diff
@@ -356,10 +376,17 @@ cpdef convert_sorted_vars_name_desc(list sorted_vars):
...
+        if i < items:
+            var = sorted_vars[i]
+            tmp_name = encode(var.name)
+            tmp_desc = encode(var.description)
+        elif i < items+nof_diag_params:
+            tmp_name = encode(diag_param_names[i-items][0])
+            tmp_desc = encode(diag_param_names[i-items][1])
+        else:
+            tmp_name = encode(diag_vars[i-items-nof_diag_params][0])
+            tmp_desc = encode(diag_vars[i-items-nof_diag_params][1])
```
jespermattsson, line 389: This looks pretty fragile. Perhaps it would be better to loop over a generator?

```suggestion
    for tmp_name, tmp_desc in itertools.chain(map(lambda var: (var.name, var.description), sorted_vars), 
                                              diag_param_names, 
                                              diag_vars):
```
It will not create any copies of the data structures involved. Requires importing itertools, though.

src/pyfmi/fmi_util.pyx
```diff
@@ -371,16 +398,17 @@ cpdef convert_sorted_vars_name_desc(list sorted_vars):
         if desc_length_trial+1 > desc_length:
              desc_length = desc_length_trial + 1
 
-    name_output = <char*>FMIL.calloc((items+1)*name_length,sizeof(char))
+
+    name_output = <char*>FMIL.calloc((items+nof_diag_params+nof_diag_vars+1)*name_length,sizeof(char))
```
jespermattsson, line 402: Considering the number of times `items+nof_diag_params+nof_diag_vars+1` is repeated, perhaps it deserves to be in a local variable?

src/pyfmi/fmi_util.pyx
```diff
@@ -392,46 +420,53 @@ cpdef convert_sorted_vars_name_desc(list sorted_vars):
...
-        var = sorted_vars[i]
-        tmp_name = encode(var.name)
+    for i in range(items+nof_diag_params+nof_diag_vars):
+        if i < items:
+            var = sorted_vars[i]
+            tmp_name = encode(var.name)
+        elif i < items+nof_diag_params:
+            tmp_name = encode(diag_param_names[i-items][0])
+        else:
+            tmp_name = encode(diag_vars[i-items-nof_diag_params][0])
```
jespermattsson, line 448: This could also use a generator expression like the one above.

jespermattsson (commented):
I did not follow everything, so I suggest a second review.

src/pyfmi/fmi_algorithm_drivers.py
agnesramle: Changed in commit below to use dicts instead.

src/pyfmi/fmi_util.pyx
agnesramle: Fixed.

src/pyfmi/fmi_util.pyx
agnesramle: Fixed

src/pyfmi/fmi_util.pyx
agnesramle: Fixed

jespermattsson:
@agnesramle Your changes after my review looks good to me. I still suggest a second review, though, since I did not understand everything.

PR #120, opened 2021-09-15
Added new option to better illustrate its purpose when enabling diagnostic data

modelonrobinandersson:
Added new option to better illustrate its purpose when enabling diagnostic data as well as tests. Did also resolve a bug with the debug file not opened correctly as well as not generating it when we are writing results to the binary result file.


src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -100,10 +100,23 @@ class AssimuloFMIAlgOptions(OptionBase):
             solver.
             Default: "Default"
 
+        dynamic_diagnostics --
+            If True, will enable logging of diagnostics data and requires that
```
agnesramle, line 104: "will enable logging of diagnostics data to the binary result file.

src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -100,10 +100,23 @@ class AssimuloFMIAlgOptions(OptionBase):
             solver.
             Default: "Default"
 
+        dynamic_diagnostics --
+            If True, will enable logging of diagnostics data and requires that
+            the option 'result_handling' is set to 'binary', otherwise an exception is raised.
+            If True, the debug file will not be generated.
```
agnesramle, line 106: All users will not now what "the debug file" refers to. I suggest to remove this line, especially with the update above.

src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -100,10 +100,23 @@ class AssimuloFMIAlgOptions(OptionBase):
             solver.
             Default: "Default"
 
+        dynamic_diagnostics --
+            If True, will enable logging of diagnostics data and requires that
+            the option 'result_handling' is set to 'binary', otherwise an exception is raised.
+            If True, the debug file will not be generated.
+            Before simulation start, a check for name clashes between
```
agnesramle, line 107: State that it is "Diagnostics" that is forbidden, maybe something in line with:

Model variables are not allowed to start with "Diagnostics* using this option. This will be checked at simulation start and an exception raised if they do.

src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -100,10 +100,23 @@ class AssimuloFMIAlgOptions(OptionBase):
...
+            The diagnostics data will be available via the simulation resuls and/or the
+            binary result file generated during simulation.
+
+            Default: False
+
         logging --
             If True, creates a logfile from the solver in the current
             directory and enables logging of diagnostics data to logfile or resultfile,
             based on simulation option 'result_handling'.
+            If True and 'result_handling' is set to 'binary', a debug file is not generated.
```
agnesramle, line 119: Do not use the name debug file, the log file from the FMU is still generated and could be mixed up with this one.

tests/test_io.py
```diff
@@ -1037,12 +1040,120 @@ def test_diagnostics_data_cancelled_simulation_file_stream(self):
...
+        # Now add test_str to file, simulate, and verify that test_str is gone
+        with open(expected_debug_file, 'w') as f:
+            f.write(test_str)
+        model.simulate(options = opts)
+
+        # Verify
+        with open(expected_debug_file, 'r') as f:
+            l = f.readline()
+        nose.tools.assert_false(test_str in l, "Test failed, found '{}' in '{}'".format(test_str, l))
+
```
agnesramle, line 1150: Cant't the me1 and me2 tests be merged to 1 with an option instead? Much code duplication here.

agnesramle (changes requested):
Good job!

PR #130, opened 2021-11-09
Updates to diagnostics

modelonrobinandersson:
* Rearranged structure of diagnostic variables.
* Reduced code duplication
* Changed prefix to avoid name conflicts with legal Modelica identifiers.
* Updated tests.


tests/test_io.py
```diff
@@ -969,7 +971,7 @@ def _get_diagnostics_cancelled_sim(self, result_file_name):
         model.initialize()
 
         # Need to mock the diagnostics variables in order to invoke simulation_start
-        diagnostics_start_name = "Diagnostics."
+        diagnostics_start_name = diagnostics_prefix
```
agnesramle, line 974: I don't think diagnostics_start_name is needed anymore. Fix if you like.

modelonrobinandersson: Good point! I've fixed it now

PR #135, opened 2022-02-20
improve handling of build depends,  PEP 518

vruge:
PEP 518: https://www.python.org/dev/peps/pep-0518/
see: https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#basic-setup-py


vruge:
no longer necessary because of #139

PR #143, opened 2022-05-11
Added functionality to allow re-reading data from file

chria:


src/common/io.py
```diff
@@ -1218,6 +1218,13 @@ def __init__(self, fname, delayed_trajectory_loading = True):
                 all at the same time. Only works for files or streams that
                 are based on a file and has attribute name.
                 Default: True (for files)
+                
+            allow_file_updates --
+                If this is turned on, file updates (in terms of more
```
modelonrobinandersson, line 1223: I would rephrase "If this is turned on" to "If this is set to True" to clarify also that it is a boolean value.

src/common/io.py
```diff
@@ -1218,6 +1218,13 @@ def __init__(self, fname, delayed_trajectory_loading = True):
...
+                data point being added to the result file) is allowed.
```
modelonrobinandersson, line 1224: I think it is missing "s" in "data point**s**"

src/common/io.py
```diff
@@ -1230,6 +1237,7 @@ def __init__(self, fname, delayed_trajectory_loading = True):
             self._fname = fname
             self._is_stream = True
             delayed_trajectory_loading = False
+        self._allow_file_updates = allow_file_updates
```
modelonrobinandersson, line 1240: Would be good to verify that allow_file_updates is a boolean here otherwise throw an exception

src/common/io.py
```diff
@@ -1274,16 +1282,42 @@ def __init__(self, fname, delayed_trajectory_loading = True):
...
+        self._data_2_info = self.raw["data_2"]
+        self._data_2 = {}
+        
+        if self._contains_diagnostic_data:
+            self._data_3_info = self.raw["data_3"]
+            self._data_3 = {}
+            self._data_4_info = self.raw["data_4"]
+            
+            self._file_pos_model_var = np.empty(self._data_2_info["nbr_points"], dtype=np.longlong)
+            self._file_pos_diag_var = np.empty(self._data_3_info.shape[0], dtype=np.longlong)
```
modelonrobinandersson, line 1313: Great that you also remembered to set dtype to np.longlong!

src/common/io.py
```diff
@@ -1324,17 +1358,20 @@ def _get_name_dict(self):
 
     def _get_trajectory(self, data_index):
         if isinstance(self._data_2, dict):
-            if data_index in self._data_2:
-                return self._data_2[data_index]
+            if self._mtime != os.path.getmtime(self._fname):
+                if self._allow_file_updates:
+                    self._update_data_info()
+                else:
+                    raise JIOError("The result file have been modified since the result object was created. Please make sure that different filenames are used for different simulations.")
```
modelonrobinandersson, line 1365: I think we should put the "raise JIOError" in a separate function now because this code is duplicated with line 1321.

src/common/io.py
```diff
@@ -1344,16 +1381,30 @@ def _get_trajectory(self, data_index):
 
     def _get_diagnostics_trajectory(self, data_index):
         """ Returns trajectory for the diagnostics variable that corresponds to index 'data_index'. """
+        if self._mtime != os.path.getmtime(self._fname):
+            if self._allow_file_updates:
+                self._update_data_info()
+            else:
+                raise JIOError("The result file have been modified since the result object was created. Please make sure that different filenames are used for different simulations.")
```
modelonrobinandersson, line 1388: Same here regarding the duplicated code as my previous comment.

src/common/io.py
```diff
@@ -1344,16 +1381,30 @@ def _get_trajectory(self, data_index):
...
     def _read_trajectory_data(self, data_index, read_diag_data):
         """ Reads corresponding trajectory data for variable with index 'data_index',
             note that 'read_diag_data' is a boolean used when this function is invoked for
             diagnostic variables.
         """
+        if self._mtime != os.path.getmtime(self._fname):
+            if self._allow_file_updates:
+                self._update_data_info()
+            else:
+                raise JIOError("The result file have been modified since the result object was created. Please make sure that different filenames are used for different simulations.")
```
modelonrobinandersson, line 1406: Same here regarding the duplicated code as my previous comment.

src/common/io.py
```diff
@@ -1376,13 +1425,23 @@ def _read_trajectory_data(self, data_index, read_diag_data):
 
     def _get_interpolated_trajectory(self, data_index):
         """ Returns an interpolated trajectory for variable of corresponding index 'data_index'. """
+        if self._mtime != os.path.getmtime(self._fname):
+            if self._allow_file_updates:
+                self._update_data_info()
+            else:
+                raise JIOError("The result file have been modified since the result object was created. Please make sure that different filenames are used for different simulations.")
```
modelonrobinandersson, line 1432: Same here regarding the duplicated code as my previous comment.

tests/test_io.py
```diff
@@ -636,6 +636,217 @@ def test_get_description(self):
...
...ostics.event_data.event_info.state_event_info.index_25', (0.0, 'Zero crossing indicator for event indicator 25')), ('@Diagnostics.event_data.event_info.state_event_info.index_26', (0.0, 'Zero crossing indicator for event indicator 26')), ('@Diagnostics.event_data.event_info.state_event_info.index_27', (0.0, 'Zero crossing indicator for event indicator 27')), ('@Diagnostics.event_data.event_info.state_event_info.index_28', (0.0, 'Zero crossing indicator for event indicator 28')), ('@Diagnostics.event_data.event_info.state_event_info.index_29', (0.0, 'Zero crossing indicator for event indicator 29')), ('@Diagnostics.event_data.event_info.state_event_info.index_30', (0.0, 'Zero crossing indicator for event indicator 30')), ('@Diagnostics.event_data.event_info.state_event_info.index_31', (0.0, 'Zero crossing indicator for event indicator 31')), ('@Diagnostics.event_data.event_info.state_event_info.index_32', (0.0, 'Zero crossing indicator for event indicator 32')), ('@Diagnostics.event_data.event_info.state_event_info.index_33', (0.0, 'Zero crossing indicator for event indicator 33')), ('@Diagnostics.event_data.event_info.event_type', (-1, 'No event=-1, state event=0, time event=1'))])
```
modelonrobinandersson, line 650: This is not necessary but some newlines here will make it easier to read

tests/test_io.py
```diff
@@ -636,6 +636,217 @@ def test_get_description(self):
         res = ResultDymolaBinary('CoupledClutches_result.mat')
 
         assert res.description[res.get_variable_index("J1.phi")] == "Absolute rotation angle of component"
+    
+    @testattr(stddist = True)
+    def test_modified_result_file_data_diagnostics(self):
```
modelonrobinandersson, line 641: Not mandatory but I prefer if we can add a docstring to the test explaining what the test does. I think it is good for maintenance. Same comment for all the tests below.

modelonrobinandersson (changes requested):
Looks good! I don't see anything urgent. I added a few comments and especially I would suggest that we add a short docstring to each test for future maintenance.

src/common/io.py
chria: Fixed. However, I believe we should allow anything that is interpreted as True (in True Python spirit : ) )

src/common/io.py
chria: Fixed!

src/common/io.py
chria: Fixed!

modelonrobinandersson (approved):
Looks great!

PR #144, opened 2022-05-13
Option to synchronize simulation with real-time

chria:


CHANGELOG
```diff
@@ -2,6 +2,8 @@
 --- PyFMI-2.9.7 ---
     * Added an argument to ResultDymolaBinary to allow for reading updated
       data from the loaded file.
+    * Added option to allow for synchronizing simulation with (scaled)
+      real-time.
```
modelonrobinandersson, line 6: Would it be more clear if we mention the name of the new option? Like "Added a new option 'synchronize_simulation' to ..."?

tests/test_fmi.py
```diff
@@ -1155,6 +1155,71 @@ def test_simulation_without_initialization(self):
...
 
         nose.tools.assert_raises(FMUException, model.simulate, options=opts)
+    
+    @testattr(stddist = True)
+    def test_simulation_with_syncronization_exception_ME(self):
+        model = Dummy_FMUModelME2([], "bouncingBall.fmu", os.path.join(file_path, "files", "FMUs", "XML", "ME2.0"), _connect_dll=False)
+        opts = model.simulate_options()
+        opts["synchronize_simulation"] = "Hej"
+        
+        nose.tools.assert_raises(FMUException, model.simulate, options=opts)
```
modelonrobinandersson, line 1165: I think it is better if we assert that it raises fmi.InvalidOptionException and not the one the exception extends. I also propose adding a docstring to the tests for maintainability.

src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -159,6 +159,18 @@ class AssimuloFMIAlgOptions(OptionBase):
             example is filter = "*der" , stor all variables ending with
             'der'. Can also be a list.
             Default: None
+            
+        synchronize_simulation --
+            If set, the simulation will be syncronized to real-time or a
+            scaled real-time, if possible. The available options are:
+                True: Simulation is syncronized to real-time
```
modelonrobinandersson, line 166: I think you accidentally dropped the letter h from the word syncronized

modelonrobinandersson (changes requested):
Looks good I have nothing major to add. Just to change the exceptions we catch in the tests, add short docstring to the tests and fix the inconsistency between syncronized and synchronized

CHANGELOG
chria: Thanks, fixed!

tests/test_fmi.py
chria: Agreed, fixed.

src/pyfmi/fmi_algorithm_drivers.py
chria: Right, fixed.

PR #145, opened 2022-05-17
Replaced have with has

modelonrobinandersson:


modelonrobinandersson:
I went ahead and merged this small fix.

PR #147, opened 2022-07-18
Avoid deprecated scipy.io.matlab.mio4 import

jschueller:


modelonrobinandersson:
Hi @jschueller, what about importing scipy followed by checking `scipy.__version__`? Using the version attribute you can strictly check if if it is older than 1.8 since the ImportError you catch in the suggested change can also occur due to other reasons.

jschueller:
why not, done

src/common/io.py
```diff
@@ -30,7 +30,12 @@
 import numpy as np
 import scipy.io
 from scipy import interpolate
-from scipy.io.matlab.mio4 import MatFile4Reader, VarReader4, convert_dtypes, mdtypes_template, mxSPARSE_CLASS
+
+from pkg_resources import parse_version
```
modelonrobinandersson, line 34: I don't think we should use `pkg_resources` because then we are introducing dependency on another python package and we would like to avoid that. Only alternate solution I see is to verify that each element from `scipy.__version__.split('.')`

modelonrobinandersson (changes requested):

src/common/io.py
jschueller: ok

modelonrobinandersson (approved):

src/common/io.py
modelonrobinandersson: Looks good!

modelonrobinandersson (approved):
Looks good!

PR #149, opened 2022-07-19
Dev pm 4739

PeterMeisrimelModelon:


CHANGELOG
```diff
@@ -1,4 +1,7 @@
 --- CHANGELOG ---
+--- PyFMI-2.10.0 ---
+    * Changed default behavior for Radau5ODE to switch to analytical and sparse Jacobians, dependent on problem size.
```
modelonrobinandersson, line 3: Perhaps we should mention anything about when this is switching, reading this immediately asks the question "when does it switch"?

doc/sphinx/source/changelog.rst
```diff
@@ -2,6 +2,9 @@
 ==========
 Changelog
 ==========
+--- PyFMI-2.10.0 ---
```
modelonrobinandersson, line 5: Same here.

setup.cfg
```diff
@@ -11,4 +11,4 @@ install_requires =
     nose >= 1.3.7
     lxml >= 4.4.1
     matplotlib > 3
-    assimulo >= 3.2
\ No newline at end of file
+    assimulo >= 3.3
```
modelonrobinandersson, line 14: Good! Perhaps we should even update README such that the dependency on Assimulo is directly visible on the front page? Or we can create a separate issue for that.

modelonrobinandersson (approved):
Looks good, see my optional comment about the changelog.

PR #155, opened 2022-09-16
Dev pm 5340, fixed bouncing ball example

PeterMeisrimelModelon:
load_fmu change removed a deprecation warning.
rtol was undefined

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,11 +4,11 @@
       data from the loaded file.
     * Added option "synchronize_simulation" to allow for synchronizing 
       simulation with (scaled) real-time.
-    * Bugfix in fmi20_bouncing_ball_native.py, would trigger a step_event every timestep.
+    * Resolved a bug in the example fmi20_bouncing_ball_native where a step-event is triggered in every timestep.
 
 --- PyFMI-2.9.6 ---
     * Added setup.cfg that lists all Python package dependencies in order to run PyFMI.
-    * Resolved an issue that would occurr when reading large result files or streams causing the data to be corrupt due to an integer overflow.
+    * Resolved an issue that would occur when reading large result files or streams causing the data to be corrupt due to an integer overflow.
 
 --- PyFMI-2.9.5 ---
     * Updated structure of diagnostics data and renamed several of the variables. The top level prefix has also been changed from Diagnostics to @Diagnostics, hence the error check for name clashes has been removed.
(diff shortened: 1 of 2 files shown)


CHANGELOG
```diff
@@ -4,6 +4,7 @@
       data from the loaded file.
     * Added option "synchronize_simulation" to allow for synchronizing 
       simulation with (scaled) real-time.
+    * Bugfix in fmi20_bouncing_ball_native.py, would trigger a step_event every timestep.
```
modelonrobinandersson, line 7: I might be a bit pedantic here but I would rephrase _bugfix_ to something else, such as `Resolved a bug in the example fmi20_bouncing_ball_native where a time step is triggered for every step event.`

src/pyfmi/examples/fmi20_bouncing_ball_native.py
```diff
@@ -104,11 +104,13 @@ def run_demo(with_plots=True):
...
         #Inform the model about an accepted step and check for step events
-        step_event = bouncing_fmu.completed_integrator_step()
+        step_event, terminate = bouncing_fmu.completed_integrator_step()
+        if terminate: 
+            return
         
         #Check for time and state events
         time_event  = abs(time-Tnext) <= 1.e-10
-        state_event = True if True in ((event_ind_new>0.0) != (event_ind>0.0)) else False
+        state_event = True in ((event_ind_new>0.0) != (event_ind>0.0))
```
modelonrobinandersson, line 113: I might be missing something but isn't it enough to write `state_event = (event_ind_new>0.0) != (event_ind>0.0)`?

modelonrobinandersson (changes requested):
Looks good, just two minor comments

CHANGELOG
PeterMeisrimelModelon: I like that suggestion, changed it to:

`Resolved a bug in the example fmi20_bouncing_ball_native where a step-event is triggered in every timestep.`

src/pyfmi/examples/fmi20_bouncing_ball_native.py
PeterMeisrimelModelon: In general, no, since event_ind(_new) is an array of all event indicators and this checks if any changed sign.

It would work in this particular example though, since there is only a single event indicator. I'd still keep the general one though.

modelonrobinandersson: I see, thank you for the clarification!

modelonrobinandersson (approved):
Looks great!

PR #166, opened 2023-02-22
Use nominals from nominal_continuous_states in Jacobian

PeterMeisrimelModelon:


src/pyfmi/fmi.pyx
```diff
@@ -8170,20 +8170,28 @@ cdef class FMUModelME2(FMUModelBase2):
...
 
             for i in range(len_v):
-                eps_pt[i] = RUROUND*(max(abs(v_pt[i]), nominals[i]))
+                eps_pt[i] = RUROUND*(max(abs(v_pt[i]), nominals_pt[i]))
         else:
             for i in range(len_v):
-                nominal = self.get_variable_nominal(valueref = v_ref_pt[i])
-                eps_pt[i] = RUROUND*(max(abs(v_pt[i]), nominal))
+                tmp_nominal = self.get_variable_nominal(valueref = v_ref_pt[i])
+                eps_pt[i] = RUROUND*(max(abs(v_pt[i]), tmp_nominal))
```
modelonrobinandersson, line 8194:
```suggestion
                eps_pt[i] = RUROUND*(
                    max(
                        abs(v_pt[i]), self.get_variable_nominal(valueref = v_ref_pt[i])
                        )
                    )
```
Is tmp_nominal only added for readability? If yes then I woud propose simply changing line 8194 by introducing some line breaks, not necessarily as above but some kind of line break could help here.

chria: Thansk for the feedback! I view this as a mathematical expression (not specifically different function invokations) and thus I believe it makes sense to have it on one line.

PR #167, opened 2023-02-23
Fix for ExplicitEuler + dynamic_diagnostics with events

PeterMeisrimelModelon:


modelonrobinandersson (approved):
This looks good, my only nitpick is to add some line breaks for readability since these lines are longer than 120 characters (I didnt count all of them but 1). But consider this as optional.

PeterMeisrimelModelon:
> This looks good, my only nitpick is to add some line breaks for readability since these lines are longer than 120 characters (I didnt count all of them but 1). But consider this as optional.

Good point. Fixed it for the changes here.

PR #173, opened 2023-03-24
Recalculate absolute tolerance when depends on preinit nominal_continuous_states

filip-stenstrom:
A common workflow in PyFMI is that absolute tolerances are calculated with model.nominal_continuous_states, but that it's set before initialization. This is incorrect when the nominals are calculated after initialization.

This PR adds a heuristic that automatically tries to recalculate atol with state nominals from after initialization.

### DIFF
--- a/src/pyfmi/fmi.pyx
+++ b/src/pyfmi/fmi.pyx
@@ -1365,6 +1365,9 @@ cdef class FMUModelBase(ModelBase):
         self.callbacks.logger  = importlogger
         self.callbacks.context = <void*>self #Class loggger
 
+        # State nominals retrieved before initialization
+        self._pre_init_nominal_continuous_states = None
+
         if log_level >= FMIL.jm_log_level_nothing and log_level <= FMIL.jm_log_level_all:
             if log_level == FMIL.jm_log_level_nothing:
                 enable_logging = False
@@ -3476,6 +3479,11 @@ cdef class FMUModelME1(FMUModelBase):
                     logging.warning(f"The nominal value for {xname} is <0.0 which is illegal according " + \
                                     f"to the FMI specification. Setting the nominal to abs({xn[i]}).")
                 xn[i] = abs(xn[i])
+
+        # If called before initialization, save values in order to later perform auto-correction
+        if self._allocated_fmu == 0:
+            self._pre_init_nominal_continuous_states = xn
+
         return xn
 
     nominal_continuous_states = property(_get_nominal_continuous_states, doc =
@@ -8041,6 +8049,11 @@ cdef class FMUModelME2(FMUModelBase2):
                     logging.warning(f"The nominal value for {xnames[i]} is <0.0 which is illegal according " + \
                                     f"to the FMI specification. Setting the nominal to abs({xn[i]}).")
                 xn[i] = abs(xn[i])
+
+        # If called before initialization, save values in order to later perform auto-correction
+        if self._initialized_fmu == 0:
+            self._pre_init_nominal_continuous_states = xn
+
         return xn
 
     nominal_continuous_states = property(_get_nominal_continuous_states, doc =
@@ -8508,10 +8521,17 @@ cdef class __ForTestingFMUModelME2(FMUModelME2):
         return FMIL.fmi2_status_ok
 
     cdef int __get_nominal_continuous_states(self, FMIL.fmi2_real_t* xnominal, size_t nx):
-        # Set some illegal values in order to test the fallback/auto-correction.
         for i in range(nx):
-            xnominal[i] = (((<int> i) % 3) - 1) * 2.0  # -2.0, 0.0, 2.0, <repeat>
+            if self._initialized_fmu == 1:
+                # Set new values to test that atol gets auto-corrected.
+                xnominal[i] = 3.0
+            else:
+                # Set some illegal values in order to test the fallback/auto-correction.
+                xnominal[i] = (((<int> i) % 3) - 1) * 2.0  # -2.0, 0.0, 2.0, <repeat>
         return FMIL.fmi2_status_ok
+
+    cpdef set_initialized_fmu(self, int value):
+        self._initialized_fmu = value
 
 def _handle_load_fmu_exception(fmu, log_data):
     for log in log_data:
--- a/src/pyfmi/fmi.pyx
+++ b/src/pyfmi/fmi.pyx
@@ -1365,9 +1365,6 @@ cdef class FMUModelBase(ModelBase):
         self.callbacks.logger  = importlogger
         self.callbacks.context = <void*>self #Class loggger
 
-        # State nominals retrieved before initialization
-        self._pre_init_nominal_continuous_states = None
-
         if log_level >= FMIL.jm_log_level_nothing and log_level <= FMIL.jm_log_level_all:
             if log_level == FMIL.jm_log_level_nothing:
                 enable_logging = False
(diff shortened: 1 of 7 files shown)


tests/test_util.py
```diff
@@ -195,19 +209,31 @@ def set_integer(self, vref, values):
...
+    # -- Test options --
+    
+    # If true, makes use of the real __ForTesting implementation for nominal_continuous_states,
+    # else just returns 1.0 for each.
+    override_nominal_continuous_states = True
+
+    # Values for nominal_continuous_states. 'test_sparse_option()' and more tests will break
+    # if the values are not calculated from get_ode_sizes() as defined at __init__.
+    # REVIEW: Can we fix the tests instead? I don't know how to.
+    _nominal_continuous_states = None
```
filip-stenstrom, line 222: Please comment if you know how that tests work. Otherwise I'll just leave this as is, since from what I can tell, it's the old tests that does something wrong.

src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -605,6 +596,43 @@ def _set_options(self):
...
+                else:
+                    self.solver_options["atol"] = 0.01*rtol*self.model.nominal_continuous_states
+            elif isinstance(preinit_nominals, N.ndarray) and (preinit_nominals.size > 0):
+                # Heuristic:
+                # Try to find if atol was specified as "atol = factor * model.nominal_continuous_states",
+                # and if that's the case, recompute atol with nominals from after initialization.
+                factors = atol / preinit_nominals
+                f0 = factors[0]
+                for f in factors:
+                    if abs(f0 - f) > f0 * 1e-6:  # REVIEW: How to do this properly?
```
filip-stenstrom, line 626: Please comment. I.e. what tolerance (rel/abs) should we use, and what value?

src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -605,6 +596,43 @@ def _set_options(self):
...
+                        # REVIEW: Should this give warning? In Case 2 we mention something about it, but hard to understand.
```
filip-stenstrom, line 627: Please comment. Case 2 as in the internal onenote.

src/pyfmi/fmi.pyx
```diff
@@ -3442,17 +3445,46 @@ cdef class FMUModelME1(FMUModelBase):
...
+        cdef N.ndarray[FMIL.fmi1_real_t, ndim=1, mode='c'] xn = N.zeros(self._nContinuousStates, dtype=N.double)
 
+        status = self.__get_nominal_continuous_states(<FMIL.fmi1_real_t*> xn.data, self._nContinuousStates)
         if status != 0:
             raise FMUException('Failed to get the nominal values.')
 
-        return ndx
+        # Fallback - auto-correct the illegal nominal values:
+        xvrs = self.get_state_value_references()
+        # TODO: What about xvrs[i] == fmiUndefinedValueReference?
```
filip-stenstrom, line 3468: Something we need to care about? I've never seen this before

chria: Can this happen? If it can I suggest that we check against it and return "state variable <index>" instead. I also have not seen this.

src/pyfmi/fmi.pyx
```diff
@@ -7799,6 +7885,8 @@ cdef class FMUModelME2(FMUModelBase2):
         the FMI specification, atol = 0.01*rtol*(nominal values of the
         continuous states).
 
+        This method should not be called before initialization, since it depends on state nominals.
```
chria, line 7888: Also here: "... on state nominals which can change during initialization"

src/pyfmi/fmi_algorithm_drivers.py
chria: I think this looks good as is.

src/pyfmi/fmi_algorithm_drivers.py
chria: Lets skip a warning.

src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -605,6 +596,43 @@ def _set_options(self):
...
+                        return
+                # Success.
+                self.solver_options["atol"] = atol * self.model.nominal_continuous_states / preinit_nominals
```
chria, line 630: I'm a bit confused why this is indented to be outside of the if-statement? Should this not be part of the ifelse block above? It seems like we do this update now even if "Default" was used or am I missing something?

tests/test_util.py
chria: Lets leave it.

src/pyfmi/fmi_algorithm_drivers.py
filip-stenstrom: I think it is part of the elif block?

src/pyfmi/fmi.pyx
filip-stenstrom: Fixed

src/pyfmi/fmi_algorithm_drivers.py
chria: You are right, it is. I missed one indentation when looking at it from the review pane.

src/pyfmi/fmi.pyx
```diff
@@ -3531,11 +3566,40 @@ cdef class FMUModelME1(FMUModelBase):
...
+        continuous states)
+
+        This method should not be called before initialization, since it depends on state nominals.
+
+        Returns::
+
+            atol --
+                The absolute tolerances.
+        """
+        rtol = self.get_relative_tolerance()
```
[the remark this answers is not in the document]
filip-stenstrom: We do, but in the FMU:
```
    if (!jmi->is_initialized) {
        jmi_log_node(jmi->log, logWarning, "GetNominals", "Getting state nominals before initialization.");
    }
```

src/pyfmi/fmi.pyx
filip-stenstrom: It's removed in FMI2 and after looking at it a bit more I don't even understand how it would work in the first place. It's defined in the C header as unsigned -1, which is platform and compiler dependent on the size of unsigned int. 

It's probably very unlikely that we need to check for it, given that most tools want to compile for both ME1 and ME2, and it doesn't make sense to have special behavior for FMI1.

We'd also get hit by it in other places.

I suggest just removing the TODO.

tests/test_fmi.py
```diff
@@ -16,21 +16,31 @@
...
-import pyfmi.fmi_algorithm_drivers as fmi_algorithm_drivers
+from pyfmi.fmi_algorithm_drivers import AssimuloFMIAlg, AssimuloFMIAlgOptions, \
+                                        PYFMI_JACOBIAN_LIMIT, PYFMI_JACOBIAN_SPARSE_SIZE_LIMIT
 from pyfmi.tests.test_util import Dummy_FMUModelCS1, Dummy_FMUModelME1, Dummy_FMUModelME2, Dummy_FMUModelCS2, get_examples_folder
 from pyfmi.common.io import ResultHandler
 from pyfmi.common.algorithm_drivers import UnrecognizedOptionError
 from pyfmi.common.core import create_temp_dir
 
+
+class TempAlg(AssimuloFMIAlg):
```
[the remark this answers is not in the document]
filip-stenstrom: Fixed

src/pyfmi/fmi.pyx
filip-stenstrom: (I removed the TODO already, but thread is still open for discussion)

PR #175, opened 2023-04-14
Updated docstring for completed_integrator_step

### LINKED ISSUE #174: completed_integrator_step returns a tuple instead of a boolean
Dear All,

I am simulating an FMU based on the example [fmi_bouncing_ball_native.py](https://github.com/modelon-community/PyFMI/blob/master/src/pyfmi/examples/fmi_bouncing_ball_native.py). 

In my implementation, however, the method `completed_integrator_step()` of an `FMUModelME2` instance, returns a tuple with two elements ([False, False]) instead of a boolean. It does not seem to be correct according to the method help text shown below:

>Help on built-in function completed_integrator_step:
>
>completed_integrator_step(...) method of pyfmi.fmi.FMUModelME2 instance
>    This method must be called by the environment after every completed step
>    of the integrator. If the return is True, then the environment must call
>    event_update() otherwise, no action is needed.
>    
>    Returns::
>    
>        True -> Call event_update().
>        False -> Do nothing.
>    
>    Calls the low-level FMI function: fmi2CompletedIntegratorStep.

Could anyone please help me on this?
Thanks in advance for the attention.

modelonrobinandersson:
This closes #174

### DIFF
--- a/src/pyfmi/fmi.pyx
+++ b/src/pyfmi/fmi.pyx
@@ -2573,7 +2573,7 @@ cdef class FMUModelBase(ModelBase):
         Returns::
 
             A dict consisting of the alias variables along with no alias variable.
-            The values indicates wheter or not the variable should be negated or not.
+            The values indicate whether or not the variable should be negated or not.
 
         Raises::
 
@@ -5033,7 +5033,7 @@ cdef class FMUModelBase2(ModelBase):
         Returns::
 
             A dict consisting of the alias variables along with no alias variable.
-            The values indicates whether or not the variable should be negated or not.
+            The values indicate whether or not the variable should be negated or not.
 
         Raises::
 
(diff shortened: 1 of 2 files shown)


src/pyfmi/fmi.pyx
```diff
@@ -7956,9 +7956,11 @@ cdef class FMUModelME2(FMUModelBase2):
         event_update() otherwise, no action is needed.
 
         Returns::
-
-            True -> Call event_update().
-            False -> Do nothing.
+            A tuple of format (a, b) where a and b indicates:
```
PeterMeisrimelModelon, line 7959: indicate

modelonrobinandersson: Good catch, I found more errors. See my last two commits.

PeterMeisrimelModelon: Looks good!

PeterMeisrimelModelon (approved):

PR #181, opened 2023-05-16
Fixed so that hasattr works on a log node

chria:


CHANGELOG
```diff
@@ -1,6 +1,7 @@
 --- CHANGELOG ---
 --- PyFMI-2.10.4 ---
     * Added 'result_handling' = None as option, deprecated 'none'.
+    * Fixed so that 'hasattr' works on log nodes.
```
PeterMeisrimelModelon, line 4: Changes look good to me, but "fixed" sounds a bit off. Unless I am missing something, it looks like it did work before, but didn't have appropriate error handling, since missing attributes would generate KeyErrors rather than AttributeError . (Or is this a requirement for the parsing to work properly?)

chria: If you have this:

class A:
   pass

a = A()

hasattr(a, "attr")

This should result in that "False" is returned. It should not throw an exception. If "A" now was a node class, hasattr(a, "attr") would instead throw an exception which is not the expected behaviour. "hasattr" expects that an AttributeError is raised whenever an attribute does not exists on a class and then automatically catches this and return False in these cases

PeterMeisrimelModelon: Ah, ok, thanks for the clarification!

PeterMeisrimelModelon (approved):

PR #186, opened 2023-06-06
Add CI script

jschueller:
similar to https://github.com/modelon-community/Assimulo/pull/39

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,57 @@
+on:
+  push:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  linux:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install system
+      run: |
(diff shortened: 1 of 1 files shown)


PeterMeisrimelModelon:
Looks good to me. 

Do you think you could also give this the same treatment of testing multiple Sundials versions as currently in Assimulo? See also https://github.com/modelon-community/Assimulo/pull/61 (I'm trying to get this a bit prettier for the extra if based on different versions, but haven't had time to look deeper into the syntax yet.)

jschueller:
I think its not worth the trouble because PyFMI does not use sundials direcly.

jschueller:
Dont you want to contribute your changes to upstream sundials instead of forking ?

PeterMeisrimelModelon:
We want to keep the sundials fork, some changes are minor customizations rather than added features. 

And I think you are right that testing that requires specific Sundials versions should be contained to Assimulo and PyFMI functionality should be largely independent. 

I'd propose to change the used Sundials version to 2.7.0 though, since it is the version we aim to be compatible with in Assimulo development.

jschueller:
ok, done

PR #185, opened 2023-06-05
Avoid fmi2_import_get_derivatives with NULL array

jschueller:
Closes #184, similar fix as in #92

Do you know why _nContinuousStates is zero in ME mode ?

cc @modelonrobinandersson

PR #187, opened 2023-06-19
Don't call FMU when calling get_derivatives when there are no states.

PeterMeisrimelModelon:
See also: https://github.com/modelon-community/PyFMI/pull/185


jschueller (approved):
great, cant wait to have this!

PR #190, opened 2023-06-21
Consistent line endings

jschueller:
Some files use unix line endings, some windows line endings (sometimes mix them the same file), and that made patching difficult in the past, fixed with:
```
$ find tests src/ -type f| xargs dos2unix
```
On windows git will handle unix lines nicely and checkout with dos line endings.

cc @modelonrobinandersson

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+* text=auto eol=lf
+


modelonrobinandersson:
Hi @jschueller thank you for the PR.

Would this PR be necessary if we instead add a file `.gitattributes` at the top with contents `* text=auto eol=lf`?

jschueller:
we can add a .gitattributes so that new files keep being converted to lf, but we still have to convert the old ones

PR #196, opened 2023-08-17
Remove FMUX and JMU related code

filip-stenstrom:
The FMUX and JMU model formats are obsolete. This PR removes related code.


src/common/xmlparser.py
```diff
@@ -1,5718 +0,0 @@
-#!/usr/bin/env python
```
PeterMeisrimelModelon, line 1: If we remove this file, the reference to it in src/common/__init__.py + references in the documentation generation should also be removed.

filip-stenstrom: Will do. As part of building it's apparently copied to pyjmi.common.xmlparser and pymodelica.common.xmlparser, so I might need to revert this tho.

src/common/core.py
```diff
@@ -449,7 +449,7 @@ def get_unit_name(class_name, unit_type='FMU'):
             The name of the model.
             
         unit_type --
```
PeterMeisrimelModelon, line 451: It looks like one could actually remove the whole `get_unit_name `function, couldn't find any place it is used.

PeterMeisrimelModelon (approved):
Looks great, happy that we can remove some obsolete parts.

PR #213, opened 2024-01-24
Upgrade to Python3.11 + Cython3

PeterMeisrimelModelon:
Still some cleanup to do, but will do that as part of another PR. This is the current minimum to get PyFMI up and running with Cython3.

Core changes are as follows/address the following issues:
1. New import path formats
2. Removing replacing "__" prefixes from variables/functions as appropriate, where not private variables.
3. Replaced <numy.ndarray>.size by numpy.size(<numpy.ndarray>). This caused segfaults, not sure why.
4. Renaming variables using built-in names; e.g., sum, type. More cases remaining.
5. Intermediate variables to cast `sqrt `output to double.
6. Cython3/newer numpy is more strict on disallowing nested arrays; various fixes related to that.
7. Some misc. cleanup.


tests/test_util.py
```diff
@@ -15,18 +15,27 @@
...
-                      FMUModelCS2, FMUModelME2, __ForTestingFMUModelME2
+from pyfmi.fmi import (FMUException, FMUModelME1, _ForTestingFMUModelME1, FMUModelCS1, 
+                      FMUModelCS2, FMUModelME2, _ForTestingFMUModelME2)
 
 def get_examples_folder():
     return os.path.join(os.path.dirname(__file__), '..', 'examples')
 
-class Dummy_FMUModelME1(__ForTestingFMUModelME1):
-    # If true, makes use of the real __ForTesting implementation for nominal_continuous_states,
+class _ValuesDict(dict):
```
PeterMeisrimelModelon, line 27: Not super happy with this.

The core issue is that e.g., Dummy_FMUModelX.values could contain values that are e.g., np.array([1.]). 
This may be isolated to _ForTestingFMUModelX but goes back to various callbacks in the Test-problem definitions. However, fixing it there would be quite messy and non-instructive.

setup.cfg
```diff
@@ -1,13 +1,13 @@
 [options]
 setup_requires =
     setuptools
-    numpy >= 1.17
-    cython >= 0.29.13
+    numpy >= 1.26.3
+    cython >= 3.0.7
```
modelonrobinandersson, line 5:
```suggestion
    cython >= 3.0
```

.github/workflows/build.yml
```diff
@@ -35,9 +40,8 @@ jobs:
         sudo make install
     - name: Install assimulo
       run: |
-        cd /tmp
-        curl -fSsL https://github.com/modelon-community/Assimulo/archive/Assimulo-3.4.1.tar.gz | tar xz
-        cd Assimulo-Assimulo-3.4.1
+        git clone --depth 1 -b dev-pm-python_upgrade https://github.com/modelon-community/Assimulo /tmp/Assimulo
+        cd /tmp/Assimulo
```
modelonrobinandersson, line 44: Should we change this git clone command before merging?

src/pyfmi/fmi.pxd
```diff
@@ -116,7 +116,7 @@ cdef class FMUModelBase(ModelBase):
     cdef FMIL.jm_callbacks* callbacks_standard
 
     #Internal values
-    cdef public object __t
+    cdef public object _t
```
modelonrobinandersson, line 119: Should we move this declaration to `ModelBase` since it is also added again on line 194? Feel free to ignore but looks like a possible clean-up

src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -1323,9 +1323,10 @@ def parameter_estimation_callback(y):
         p0 = []
         for i,parameter in enumerate(self.parameters):
             p0.append(self.model.get(parameter)/self.options["scaling"][i])
+        p0 = N.array(p0).flatten()
```
modelonrobinandersson, line 1326: Doesn't this change affect the line 1331 further down? Since now we give `sciopt.minimize` a flattened array but previously we did not, maybe it doesn't matter?

tests/test_util.py
modelonrobinandersson: Do we need a ticket to clean this up? I do agree it looks messy I wonder if we can improve it somehow, but then as a separate effort.

tests/test_util.py
```diff
@@ -286,8 +284,9 @@ def completed_integrator_step(self, *args, **kwargs):
     def get_derivatives(self):
         return -self.continuous_states
 
-    def get_real(self, vref):
-        self.get_derivatives()
+    def get_real(self, vref, evaluate = True):
```
PeterMeisrimelModelon, line 287: See comment in tests

tests/test_fmi_estimate.py
```diff
@@ -101,11 +97,11 @@ def f(*args, **kwargs):
             model.reset()
 
             # Set optimal values for a1 and a2 into the model
-            model.set('qt.a1',res["qt.a1"])
-            model.set('qt.a2',res["qt.a2"])
+            model.set(['qt.a1'], res["qt.a1"])
```
PeterMeisrimelModelon, line 100: Simply used incorrectly, but was accepted in the DummyModel anyways and broke array concatination in various places, since new Cython is more strict on this.

modelonrobinandersson (approved):

PR #214, opened 2024-01-25
Cleaning up various Cython warnings

PeterMeisrimelModelon:
Some warnings that remain, but not worthwhile fixing:

1. `IF` deprecation from Cython, no proper alternative exists thus far.
2. Comparisons of different signedness, originating from e.g., np.array([0.]*nref), where nref is size_t (unsigned). This would become a loop in the generated C code, with an ssize_t as running index, which would be signed and the comparison with nref would give the warning. See also https://peps.python.org/pep-0353/
3. Not catching returns from e.g., fread(...)
4. An unused variables that is generated from the C code, but (not obviously) fixable on the Cython side.

### DIFF
--- a/src/pyfmi/fmi.pyx
+++ b/src/pyfmi/fmi.pyx
@@ -130,7 +130,7 @@ FMI2_KIND_DISCRETE  = FMIL.fmi2_dependency_factor_kind_discrete
 FMI_ME                 = FMIL.fmi1_fmu_kind_enu_me
 FMI_CS_STANDALONE      = FMIL.fmi1_fmu_kind_enu_cs_standalone
 FMI_CS_TOOL            = FMIL.fmi1_fmu_kind_enu_cs_tool
-FMI_MIME_CS_STANDALONE = encode("application/x-fmu-sharedlibrary")
+cdef FMI_MIME_CS_STANDALONE = encode("application/x-fmu-sharedlibrary")
 
 FMI_REGISTER_GLOBALLY = 1
 FMI_DEFAULT_LOG_LEVEL = FMIL.jm_log_level_error
@@ -1501,7 +1501,7 @@ cdef class FMUModelBase(ModelBase):
 
             model.get_version()
         """
-        version = FMIL.fmi1_import_get_version(self._fmu)
+        cdef FMIL.fmi1_string_t version = <FMIL.fmi1_string_t>FMIL.fmi1_import_get_version(self._fmu)
         return decode(version)
 
     def get_ode_sizes(self):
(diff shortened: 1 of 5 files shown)


src/pyfmi/fmi.pyx
```diff
@@ -1501,7 +1501,7 @@ cdef class FMUModelBase(ModelBase):
 
             model.get_version()
         """
-        version = FMIL.fmi1_import_get_version(self._fmu)
+        cdef FMIL.fmi1_string_t version = <FMIL.fmi1_string_t>FMIL.fmi1_import_get_version(self._fmu)
```
PeterMeisrimelModelon, line 1504: This and many other places; various returns are FMIL.* strings instead of char* and the explicit typecast suppresses a warning on lost `const`

src/pyfmi/fmil_import.pxd
```diff
@@ -56,18 +55,18 @@ cdef extern from 'fmilib.h':
...
+    ctypedef void*  fmi1_component_t
+    ctypedef void*  fmi2_component_t
+    ctypedef char*  fmi1_string_t
+    ctypedef char*  fmi2_string_t
     ctypedef int    fmi1_integer_t
     ctypedef int    fmi2_integer_t
-    ctypedef long unsigned int size_t
-    ctypedef void * jm_voidp
-    ctypedef char * jm_string
+    # ctypedef long unsigned int size_t # same as via stdlib.h
```
PeterMeisrimelModelon, line 64: This gave quite a few warnings.

src/pyfmi/fmil_import.pxd
```diff
@@ -545,7 +544,7 @@ cdef extern from 'fmilib.h':
     char * fmi1_import_get_GUID(fmi1_import_t *)
 
     #FMI HELPER METHODS (2.0)
-    char * fmi2_get_types_platform()
+    # char * fmi2_get_types_platform()
```
PeterMeisrimelModelon, line 547: Unused; keeping them around though in case we want them in the future

src/pyfmi/fmil_import.pxd
modelonrobinandersson: I wonder if we should just remove it, we will always have commit history and perhaps we should just be less conservative in the clean-up approach? Feel free to ignore and resolve.

src/pyfmi/fmil_import.pxd
modelonrobinandersson: Same comment here as above.

modelonrobinandersson (approved):

src/pyfmi/fmil_import.pxd
PeterMeisrimelModelon: I think these (also others lower down) are good to keep around, to be able to see "This exists in FMIL, but is unused right now".

PR #219, opened 2024-02-05
Removing obsolete test files

PeterMeisrimelModelon:
These test files are not used in any tests.


modelonrobinandersson (approved):
I'd suggest to check also with @chria if there is anything else to these FMUs that we might not be aware of, but looks good to me!

PR #221, opened 2024-02-13
nose to pytest conversion

PeterMeisrimelModelon:


pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+filterwarnings =
+    ignore:.*does not support directional derivatives.*:UserWarning
```
PeterMeisrimelModelon, line 3: Config file for pytest, ingore all warnings matching the regex "*does not support directional derivatives.*" and of type UserWarning

src/pyfmi/__init__.py
```diff
@@ -34,20 +34,6 @@
 int = np.int32
 np.int = np.int32
 
-def testattr(**kwargs):
```
PeterMeisrimelModelon, line 37: unused except for one case; one test marked for windows, but works for linux anyways.

src/pyfmi/master.pyx
```diff
@@ -756,7 +756,7 @@ cdef class Master:
             index_end = index_start + self.models_dict[model]["local_output_len"]
             local_output_vref_array = (<FMUModelCS2>model).get_real(self.models_dict[model]["local_output_vref_array"])
             for i, index in enumerate(range(index_start, index_end)):
-                y[index] = local_output_vref_array[i]
+                y[index] = local_output_vref_array[i].item()
```
PeterMeisrimelModelon, line 759: `model.get_real `will always return a numpy array, setting a scalar with an array is deprecated.

+ a few more occasions in this file.

tests/test_io.py
```diff
@@ -1656,37 +1544,30 @@ def _work_flow_me2(self, result_file_name):
...
         """ Verify exception when using ResultHandlerCSV with a stream that doesnt support 'write'. """
         class A:
             pass
         stream = A() # send in something that is not a string
         msg = "Failed to write the result file. Option 'result_file_name' needs to be a filename or a class that supports writing to through the 'write' method."
-        with nose.tools.assert_raises_regex(FMUException, msg):
+        with pytest.raises(FMUException, match = msg):
             self._work_flow_me2(stream)
 
-    """
```
PeterMeisrimelModelon, line 1682: These were commented-out; adding them back since they do work.

tests/test_fmi.py
```diff
@@ -1550,11 +1476,10 @@ def test_simulation_with_syncronization_ME(self):
         res = model.simulate(final_time=0.1, options=opts)
         tsyn = res.detailed_timings["computing_solution"]
         
-        assert tsyn > t, "Syncronization does not work: %d, %d"%(t, tsyn)
+        assert tsyn > t, "synchronization does not work: %d, %d"%(t, tsyn)
```
modelonrobinandersson, line 1479:
```suggestion
        assert tsyn > t, f"synchronization does not work: Expected {tsyn} > {t}"
```

tests/test_fmi.py
```diff
@@ -1572,9 +1497,8 @@ def test_simulation_with_syncronization_CS(self):
...
```
modelonrobinandersson, line 1500:
```suggestion
        assert tsyn > t, f"synchronization does not work: Expected {tsyn} > {t}"
```

tests/test_io.py
```diff
@@ -130,9 +124,8 @@ def test_read_alias_derivative(self):
             dery = res["der(y)"]
```
modelonrobinandersson, line 125: Should we also add `assert len(derx) > 0` to make sure the loop actually iterates over anything?

tests/test_io.py
```diff
@@ -506,11 +469,9 @@ def test_read_alias_derivative(self):
...
 
             for i in range(len(derx)):
-                nose.tools.assert_equal(derx[i], dery[i])
+                assert derx[i] == dery[i]
```
modelonrobinandersson, line 472: Same comment here as above

modelonrobinandersson (approved):
Looks good, just minor comments.
Do all the tests pass on both Windows and Linux?

tests/test_io.py
PeterMeisrimelModelon: Good catch, yes, we should absolutely test that; adding.

tests/test_fmi_coupled.py
```diff
@@ -15,81 +15,78 @@
...
     import assimulo
 except ImportError:
-    assimulo_installed = False
+    pass
 
 file_path = os.path.dirname(os.path.abspath(__file__))
-
 me2_xml_path = os.path.join(file_path, "files", "FMUs", "XML", "ME2.0")
 
-if assimulo_installed:
```
PeterMeisrimelModelon, line 37: I think this was a fundamentally flawed approach, if one doesn't have Assimulo installed, a user should not run the tests that require Assimulo instead.

tests/test_fmi_coupled.py
```diff
@@ -15,81 +15,78 @@
...
+        assert res.final("Second.u2") == pytest.approx(-0.0011806893910324295)
+        assert res.initial("First.u1") == pytest.approx(-17.736842105263158)
+        assert res.initial("Second.u2") == pytest.approx(-14.73684210526316)
         
 
 class Test_CoupledFMUModelME2:
     def test_reversed_connections(self):
         model_sub_1 = FMUModelME2(os.path.join(me2_xml_path, "LinearStability.SubSystem1.fmu"), _connect_dll=False)
         model_sub_2 = FMUModelME2(os.path.join(me2_xml_path, "LinearStability.SubSystem2.fmu"), _connect_dll=False)
-        model_full  = FMUModelME2(os.path.join(me2_xml_path, "LinearStability.FullSystem.fmu"), _connect_dll=False)
```
PeterMeisrimelModelon, line 96: unused, same with a few more below.

src/pyfmi/master.pyx
```diff
@@ -782,7 +782,7 @@ cdef class Master:
...
```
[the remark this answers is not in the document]
PeterMeisrimelModelon: It would be an array of length 1, e.g., `[1.23]` and this type of assignment gave a warning from numpy.

`np.ndarray.item() (== np.ndarray.item(None))` converts this to a scalar and gives an Exception if it is not a scalar. I strongly prefer this approach over `y[index] = local_output_vref_array[i][0]`.

tests/test_fmi_coupled.py
```diff
@@ -15,103 +15,95 @@
...
 import pyfmi.fmi as fmi
-from pyfmi.tests.test_util import Dummy_FMUModelME2
+from pyfmi.test_util import Dummy_FMUModelME2
 
-assimulo_installed = True
 try:
     import assimulo
 except ImportError:
-    assimulo_installed = False
+    pass
```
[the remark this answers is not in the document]
PeterMeisrimelModelon: Added, also in a few more places.

PeterMeisrimelModelon:
> Looks good, just minor comments. Do all the tests pass on both Windows and Linux?

All the tests pass on both Windows and Linux.

src/pyfmi/master.pyx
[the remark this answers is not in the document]
modelonrobinandersson: I see, thanks!

PR #224, opened 2024-03-18
More DeprecationWarning fixes

PeterMeisrimelModelon:


src/pyfmi/fmi.pyx
```diff
@@ -5767,7 +5767,7 @@ cdef class FMUModelBase2(ModelBase):
         relative_quantity = 1 if relative_quantity_bool == FMI2_TRUE else 0
 
         vr = FMIL.fmi2_import_get_variable_vr(variable)
-        value = self.get_real(vr)
+        value = self.get_real(vr)[0]
```
PeterMeisrimelModelon, line 5770: Returns array, but value is  FMIL.fmi2_real_t

src/common/core.py
```diff
@@ -685,12 +693,7 @@ def eval(self, x):
             Two dimensional n x m matrix containing the ordinate values 
             corresponding to the argument x.
         """
-        try:
-            y = np.array(np.matrix(self.traj(float(x))))
```
PeterMeisrimelModelon, line 689: matrix is deprecated

src/common/core.py
```diff
@@ -685,12 +693,7 @@ def eval(self, x):
...
-            y = np.array(np.matrix(self.traj(x)).transpose())
-                                       #In order to guarantee that the
-                                       #return values are on the correct
-                                       #form. May need to be evaluated
-                                       #for speed improvements.
-        return y
+        if hasattr(x, "__iter__"):
+            return np.vstack([self._single_eval(i) for i in x])
+        else:
+            return self._single_eval(x)
```
modelonrobinandersson, line 699: is there any need to split this into a separate function of it is only invoked from here?

modelonrobinandersson (approved):

src/common/core.py
PeterMeisrimelModelon: You mean the new `_single_eval `I added? Strictly speaking, no; but I think it makes it easier to follow what is going on.

modelonrobinandersson: Is it possible then maybe to rename it? Initially when reading the code I was thinking "okay what is single eval?", and then I look in the function and I see we are simply returning `np.array(self.traj(x))` with some reshaping, so perhaps something could be clarified here? I have no good suggestion here but I believe it would help someone unfamiliar with this part as me.

PeterMeisrimelModelon: Changed the naming slightly and added a docstring, please have another look if you think this is more understandable now.

modelonrobinandersson: Looks good!

PR #227, opened 2024-04-16
[FEAT] PyFMI logging frequency

modelonrobinandersson:


src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -1026,13 +1040,19 @@ def _set_options(self):
...
+
     def _set_solver_options(self):
         """
         Helper function that sets options for the solver.
         """
         pass #No solver options
 
+    def integration_point_with_downsampling(self):
+        """ Invokes the integration_point method on the result handler but accounts for the downsampling option. """
+        if self.downsampling_counter % self.options['result_downsampling_factor'] == 0:
```
modelonrobinandersson, line 1052: Note that values
```
[1, 2, 3, 4, 5, 10, 100, 249, 250, 251, 499, 500]
```
results in 
```
[501, 251, 167, 126, 101, 51, 6, 3, 3, 2, 2, 2]
```
data points in the results.

src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -1046,7 +1066,9 @@ def solve(self):
 
         #For result writing
         start_time_point = timer()
+        # Always log the start even if we are downsampling
         result_handler.integration_point()
+        self.downsampling_counter += 1
```
modelonrobinandersson, line 1071: I always also thinking of updating this by calling the new function, with some kind of boolean argument "force_integration_point", then all the calls to integration_point will be via the new function.

src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -846,19 +842,27 @@ class FMICSAlgOptions(OptionBase):
...
+                        to one real-time second.
             Default: False
 
+        result_downsampling_factor --
+            Limits the number simulation results that are stored.
+            The start point and end point will still always be written, but all data stored
+            in-between is limited by this
+            Example: If set to 2: Write only half of the simulation data.
+                     If set to 10: Write only a tenth of the simulation data.
+            Default: 1 (no downsampling, i.e. the simulation data will be of length ncp + 1)
```
PeterMeisrimelModelon, line 864:
```suggestion
            int > 0, save solution to result every <result_downsampling_factor>-th communication point.
            Start & end point are always be included.
            Example: If set to 2: Result contains only every other communication point.
            Default: 1 (no downsampling)
```
A bit more brief

src/pyfmi/fmi_algorithm_drivers.py
PeterMeisrimelModelon: I think this should not be required; one can always use the ordinary `self.result_handler.integration_point()` for that

src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -927,6 +932,8 @@ def __init__(self,
         self.input = input
 
         self.status = 0
+        # used for keeping track of when to write data when downsampling the simulation results
+        self.downsampling_counter = 0
```
PeterMeisrimelModelon, line 936: I would call it something different, since this counter represents the # of the step we are in.

On second thought, I think it may even be easier to iterate over the grid (`for t in grid:`) using `for i, t in enumerate(grid):` instead, then `i` (maybe with better name) becomes what is now `self.downsampling_counter`.

Could even skip the extra function and directly put it into the loop in `solve`.

src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -1091,7 +1116,7 @@ def solve(self):
             final_time = t+h
 
             start_time_point = timer()
-            result_handler.integration_point()
+            self.integration_point_with_downsampling()
```
PeterMeisrimelModelon, line 1119: The final point should be stored unconditionally?

PeterMeisrimelModelon (commented):
+ Changelog entry

tests/test_fmi_master.py
```diff
@@ -394,6 +395,137 @@ def do_step2(current_t, step_size, new_step=True):
...
+        for value in test_values:
+            try:
+                self._sim_basic_simulation(models, connections, {'result_downsampling_factor': value})
+                error_raised = False
+            except FMUException as e:
+                error_raised = True
+                assert expected_substr in str(e), f"Error was {str(e)}, expected substring {expected_substr}"
+            assert error_raised
+        
+    # TODO: Test case that supports storing FMU states required
```
PeterMeisrimelModelon, line 466: I don't think we have any example that would work here readily available and I am not sure if it is worth it to construct one here just to test the warning.

src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -1091,7 +1107,9 @@ def solve(self):
             final_time = t+h
 
             start_time_point = timer()
-            result_handler.integration_point()
+            # down-sampling of result; step starts at 0
+            if ((step + 1) % self.result_downsampling_factor == 0) or ((step + 1) == self.ncp):
```
PeterMeisrimelModelon, line 1111: I effectively inlined what was previously a function, since this would be the only placed used.

tests/test_fmi_master.py
modelonrobinandersson: I agree

src/pyfmi/master.pyx
```diff
@@ -1455,8 +1478,22 @@ cdef class Master:
                 self.error_controlled = 1
                 self.atol = options["atol"]
                 self.rtol = options["rtol"]
+            if (self.error_controlled == 1) and (options["result_downsampling_factor"] != 1): # result_downsampling_factor = 1 is default
+                warnings.warn("Result downsampling not supported for error controlled simulation, no downsampling will be performed.")
         else:
             self.error_controlled = 0
+        
+        # Since isinstance(<any boolean>, int) evaluates to True
```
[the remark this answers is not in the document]
PeterMeisrimelModelon: Good point; fixed.

PR #236, opened 2024-05-17
Allow to build without distutils

jschueller:
Allows to use Python 3.12
it drops the no_msvcr and prefix options (that can be worked around with building the wheel then using pip --target-dir)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,19 +15,18 @@
 # You should have received a copy of the GNU Lesser General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
 
-import distutils
 import os
 import shutil
 import numpy as np
 import ctypes.util
 import sys
 
-#If prefix is set, we want to allow installation in a directory that is not on PYTHONPATH
-#and this is only possible with distutils, not setuptools
-if str(sys.argv[1:]).find("--prefix") == -1:
-    from setuptools import setup  
-else:
-    from distutils.core import setup
+try:
+    from numpy.distutils.core import setup
+    have_nd = True
+except ImportError:
+    from setuptools import setup
+    have_nd = False
 
 try:
     from Cython.Distutils import build_ext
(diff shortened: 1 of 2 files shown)


modelonrobinandersson:
@jschueller is this ready for review?

modelonrobinandersson:
@jschueller have you been able to run the tests locally with Python 3.12? Our github actions is still on 3.9 so just want to know what has been tested before approving. I guess this PR is also dependent on this PR https://github.com/modelon-community/Assimulo/pull/98, is that correct? Many thanks in advance!

jschueller:
yes, I tested this since this is how its patched in conda-forge, and indeed to build assimulo too you need the same in assimulo first

setup.py
```diff
@@ -178,14 +164,6 @@
                 shutil.copy2(path_gcc_lib,os.path.join(".","src","pyfmi"))
                 gcc_lib = os.path.join(".","src","pyfmi","libgcc_s_dw2-1.dll")
 
-if no_msvcr:
```
modelonrobinandersson, line 181: I'm a little bit hesitant to remove this for now as I'm not sure what the consequences are (or if this is something old and unused). Is it possible to change this for now to make sure we still support the `no_msvcr` if Python version is less than 3.12?
Other than this I think the changes look go od. I would just add a changelog entry.

jschueller: It looks that is only needed if mingw is used, do you still plan on using it ?
We dont need that for conda-forge.

modelonrobinandersson: It is still necessary for users of PyFMI on Windows, hence why we are not able to drop it at the moment.

jschueller: ok

modelonrobinandersson (approved):
I think just a changelog entry is missing and then it is good to go, approving. Thank you!

jschueller:
done

modelonrobinandersson (approved):

PeterMeisrimelModelon (approved):

jschueller:
ok to merge

PR #243, opened 2024-06-17
CI: Try scipy 1.13.x

jschueller:


jschueller:
Same here: I see no reason to pin scipy<1.13

PeterMeisrimelModelon:
It does appear that the previous issues was with scipy 1.13.0 since it has disappeared with scipy 1.13.1

PR #245, opened 2024-06-18
Try numpy 2.x

jschueller:


jschueller:
this is ready for review @modelonrobinandersson

modelonrobinandersson (approved):
I assume this has no breaking change to the older version of numpy.

jschueller:
This is backward compatible

PR #249, opened 2024-07-10
Updated error message with respect to needsExecutionTool

modelonrobinandersson:
The error message raised for FMUs that require an external execution tool could be improved

### DIFF
--- a/src/pyfmi/fmi.pyx
+++ b/src/pyfmi/fmi.pyx
@@ -7002,7 +7002,7 @@ cdef class FMUModelCS2(FMUModelBase2):
                 raise InvalidVersionException("The FMU could not be loaded. This class only supports FMI 2.0 for Co-simulation.")
 
         if self.get_capability_flags()['needsExecutionTool']:
-            raise FMUException('Models that need an execution tool are not supported')
+            raise FMUException('The FMU specifies that it requires an external execution tool to simulate, this is not supported.')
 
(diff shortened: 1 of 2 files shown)


src/pyfmi/fmi.pyx
```diff
@@ -7002,7 +7002,7 @@ cdef class FMUModelCS2(FMUModelBase2):
                 raise InvalidVersionException("The FMU could not be loaded. This class only supports FMI 2.0 for Co-simulation.")
 
         if self.get_capability_flags()['needsExecutionTool']:
-            raise FMUException('Models that need an execution tool are not supported')
+            raise FMUException('The FMU specifies that it requires an external execution tool to simulate, this is not supported.')
```
PeterMeisrimelModelon, line 7005:
```suggestion
            raise FMUException('The FMU specifies that it requires an external execution tool ("needsExecutionTool=True") to simulate, this is not supported.')
```
I think a concrete pointer to this flag could be useful as well here, then one has a starting point for googling.

modelonrobinandersson: I was considering adding that as well, I just wonder if people are aware of the different flags within the FMU.
How about the following?
"The FMU specifies 'needsExecutionTool=true' which implies that it requires an external execution tool to simulate, this is not supported."

modelonrobinandersson: Updated

PeterMeisrimelModelon (approved):

PR #252, opened 2024-07-25
Adding log truncation&checkpoints to ensure log truncation aligns with complete CAPI calls

PeterMeisrimelModelon:


src/pyfmi/fmil_import.pxd
```diff
@@ -467,64 +467,83 @@ cdef extern from 'fmilib.h':
         pass
 
     #FMI SPECIFICATION METHODS (2.0)
+    # basic
```
PeterMeisrimelModelon, line 470: Don't bother reviewing this; what has been done:

1. Removed a few commented-out functions
2. Sorted functions into CPAI vs non CAPI. There was some existing sorting, but neither complete nor correct.

src/common/log/handler.py
```diff
@@ -0,0 +1,65 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+"""
+XXX: There are some practical limitations around how this works
```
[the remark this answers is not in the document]
PeterMeisrimelModelon: Actually, thinking about it a bit more, the only issue we have is that if we exceed the maximal log size and then increase it, we would not resume logging. This can be fixed though. Everything else should work as expected.

src/pyfmi/fmi.pyx
```diff
@@ -693,8 +703,9 @@ cdef class ModelBase:
             if self._max_log_size_msg_sent:
                 return
 
-            msg = "The log file has reached its maximum size and further log messages will not be saved. To change the maximum size of the file, please use the 'set_max_log_size' method."
+            msg = "The log file has reached its maximum size and further log messages will not be saved. To change the maximum size of the file, please use the 'set_max_log_size' method.\n"
             self._max_log_size_msg_sent = True
+            self._current_log_size = self._current_log_size + len(msg)
```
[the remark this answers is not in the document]
PeterMeisrimelModelon: We _could_, I'd rather not though, it will make the code quite messy due to the various ways how one can log. I think in scenarios where writing to stream/file fails, it is acceptable to not be 100% accurate in hitting the limit.

src/pyfmi/fmi.pyx
```diff
@@ -242,6 +242,9 @@ cdef class ModelBase:
         self._modelId = None
         self._invoked_dealloc = 0 # Set to 1 when __dealloc__ is called
         self._result_file = None
+        # XXX: Needs to be here for now, due to cicular dependency
```
modelonrobinandersson, line 245:
```suggestion
        # XXX: Needs to be here for now, due to circular dependency 
```

src/pyfmi/fmi.pyx
```diff
@@ -575,6 +578,7 @@ cdef class ModelBase:
                 file_path -- path to extracted XML file
             otherwise function returns nothing
         """
+        # XXX: Needs to be here for now, due to cicular dependency, since pyfmi.common.log.parser imports FMUException (from this file)
```
modelonrobinandersson, line 581:
```suggestion
        # XXX: Needs to be here for now, due to circular dependency, since pyfmi.common.log.parser imports FMUException (from this file)
```

src/pyfmi/fmi.pyx
```diff
@@ -589,7 +593,13 @@ cdef class ModelBase:
             except AttributeError:
                 raise FMUException("In order to extract the XML-log from a stream, it needs to support 'seek'")
 
-        extract_xml_log(file_name, self._log_stream if is_stream else self.get_log_filename(), module_name)
+        # XXX: Needs to be here for now, due to cicular dependency
```
modelonrobinandersson, line 596:
```suggestion
        # XXX: Needs to be here for now, due to circular dependency 
```

src/pyfmi/fmi.pyx
```diff
@@ -728,15 +738,24 @@ cdef class ModelBase:
         if self._additional_logger:
             self._additional_logger(module, log_level, message)
 
-        full_msg = "FMIL: module = %s, log level = %d: %s\n"%(module, log_level, message)
+        if self._max_log_size_msg_sent:
+            return
+
+        msg = "FMIL: module = %s, log level = %d: %s\n"%(module, log_level, message)
+
+        if self._current_log_size + len(msg) > self._max_log_size:
```
modelonrobinandersson, line 746:
```suggestion
        if self._current_log_size + len(msg) >= self._max_log_size:
```

src/pyfmi/tests/test_util.pyx
```diff
@@ -330,7 +330,25 @@ class Dummy_FMUModelME2(_ForTestingFMUModelME2):
...
+        self._log_msg_prefix = "FMIL: module = Model, log level = 4: [INFO][FMU status:OK]"
+
         self.reset()
+    
+    def _logger(self, msg):
+        # emulating logging from a CAPI call
+        if self._with_logging:
+            if self._max_log_size_msg_sent:
+                return
+            msg = self._log_msg_prefix + " " + msg
```
modelonrobinandersson, line 346:
```suggestion
            msg = f"{self._log_msg_prefix} {msg}"
```

src/pyfmi/fmi.pyx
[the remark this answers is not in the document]
modelonrobinandersson: Sounds good to me!

src/pyfmi/fmi.pyx
PeterMeisrimelModelon: I'd actually keep this as is.

My interpretation around most language around log size limits in doc strings is that the limit is inclusive, i.e., we only stop logging after the limit is passed. This would change it to a non-inclusive limit + it would also require more adjustments in other place.

modelonrobinandersson: No problem, let's stick with your proposal.

modelonrobinandersson (approved):

PR #260, opened 2024-08-22
Fixed the custom result handler for the Master algorithm

PeterMeisrimelModelon:


src/pyfmi/master.pyx
```diff
@@ -281,15 +281,16 @@ class MasterAlgOptions(OptionBase):
 
         result_handling --
             Specifies how the result should be handled. Either stored to
-            file or stored in memory. One can also use a custom handler.
+            file or stored in memory. One can also use a custom handlers.
```
modelonrobinandersson, line 284:
```suggestion
            file or stored in memory. One can also use custom handlers.
```

src/pyfmi/master.pyx
```diff
@@ -281,15 +281,16 @@ class MasterAlgOptions(OptionBase):
...
             Available options: "file", "binary", "memory", "csv", "custom"
             Default: "binary"
 
         result_handler --
             The handler for the result. Depending on the option in
             result_handling this either defaults to ResultHandlerFile
             or ResultHandlerMemory. If result_handling custom is choosen
-            This MUST be provided.
+            This MUST be provided and be a dictionary with model name as 'key'
+            and the ResultHandler instance for that model as 'value'.
```
modelonrobinandersson, line 293: Just food for thought, but we could also change the API here to instead force the user to use a class. Let's say for example this class is called `Foo`, and we can say (in the docstring above) "This must be provided as an instance of `pyfmi.common.io.Foo` ...".
And the class 'Foo' is a simple class that just takes the arguments `Foo(model, ResultHandler)`. The user does not do anything else with `Foo` other than just setting it as the value for `ResultHandler`, this could also simplify the error checking below.

src/pyfmi/master.pyx
```diff
@@ -1175,9 +1176,20 @@ cdef class Master:
...
-            result_object = get_result_handler(model, opts)
+        if (opts["result_handling"] == "custom") and (not isinstance(opts["result_handler"], dict)):
+            raise fmi.FMUException("'result_handler' option must be a dictionary for 'result_handling' = 'custom'.")
+
+        for i, model in enumerate(self.models_dict.keys()):
+            if opts["result_handling"] == "custom":
+                try:
+                    handler = opts["result_handler"][model]
+                except KeyError:
+                    raise fmi.FMUException("'result_handler' option does not contain result handler for model '{}'".format(model))
```
modelonrobinandersson, line 1187: Does `model` here in the error message result in the name of the FMU?

modelonrobinandersson (approved):
I left a comment but I am approving the changes (feel free to ignore my suggestion), I think it looks good. The comment was just in case you think it makes more sense as a user.

src/pyfmi/master.pyx
PeterMeisrimelModelon: While one could, I do however not really see the advantage. Dictionaries are perfectly fit for the purpose here and consistent with the inputs we require in the `result_file_name` option.

modelonrobinandersson: My initial comment was more about that we tend to give the user a lot of freedom for inputs, and then we have to do a lot of error handling to check the types that they could potentially set it to.
Feel free to close this thread, I agree that there is little to no gain here in this scenario. It was more just a thought I had while reading through the changes 😄

src/pyfmi/master.pyx
modelonrobinandersson: Looks great!

modelonrobinandersson (approved):

PR #267, opened 2024-09-16
Added new method get_variables_data

modelonrobinandersson:


src/common/io.py
```diff
@@ -1477,6 +1475,125 @@ def get_variable_data(self, name):
...
+            Parameters::
+
+                names --
+                    List of variables names for which to fetch trajectories.
+
+                start_index --
+                    (default: 0) Starting index for trajectory slicing.
+
+                stop_index --
+                    (default: None) Stopping index for trajectory slicing;
```
modelonrobinandersson, line 1496: Thinking of removing `(default: None)` since it is obvious from the type annotations now.

src/common/io.py
```diff
@@ -1477,6 +1475,125 @@ def get_variable_data(self, name):
...
+                    name == f'{DIAGNOSTICS_PREFIX}cpu_time'
+                ):
+                trajectories.append(
+                    Trajectory(
+                        tt, self.get_variable_data(f'{DIAGNOSTICS_PREFIX}cpu_time_per_step').x[start_index:stop_index]
+                    )
+                )
+                continue
+            else:
+                # expect potential issues with diagnostics data right here
```
modelonrobinandersson, line 1561: I am removing this line `# expect potential issues...` since it is redundant now.

src/common/io.py
```diff
@@ -1477,6 +1475,123 @@ def get_variable_data(self, name):
...
+                stop_index --
+                    Stopping index for trajectory slicing.
+
+            Returns::
+                Tuple: (List of trajectories, next start index (non-negative))
+        """
+
+        """
+            TODO:
+            For [start_index, stop_index] we can do performance improvements
```
chria, line 1506: I think we should include this performance optimization within this ticket.

src/common/io.py
```diff
@@ -1477,6 +1475,123 @@ def get_variable_data(self, name):
...
+            since methods such as _get_trajectory, get_diagnostics_trajectory
+            and _get_interpolated_trajectory always retrieve the full
+            trajectory which is then sliced, here. Instead we can account for
+            start_index and stop_index and reduce the time spent reading unused data points.
+        """
+        trajectories = []
+
```
chria, line 1513: Check that stop_index is > than start_index?

src/common/io.py
```diff
@@ -1477,6 +1475,123 @@ def get_variable_data(self, name):
...
+        # Need to account for data that might be added while we are iterating over 'names' later
+        if stop_index is None:
+            stop_index = len(tt) + start_index
+
+        for name in names:
+            if isinstance(name, bytes):
+                name = decode(name)
+
+            if name == 'time' or name == 'Time':
```
chria, line 1529: As all the branches in the if-statement have "continue" - then there is no need to make is nested. Keep it simple by just writing:

if name:
   ...
   continue

if diag:
   ...
   continue

Makes it easier to read.

src/common/io.py
```diff
@@ -1477,6 +1475,123 @@ def get_variable_data(self, name):
...
+                trajectories.append(Trajectory(tt, tt))
+                continue
+            elif self._contains_diagnostic_data and (
```
chria, line 1532: How about refactor this and use a common approach both here and in get_variable_data?

src/common/io.py
```diff
@@ -1477,6 +1475,123 @@ def get_variable_data(self, name):
...
+                trajectories.append(
+                    Trajectory(
+                        tt, self.get_variable_data(f'{DIAGNOSTICS_PREFIX}cpu_time_per_step').x[start_index:stop_index]
+                    )
+                )
+                continue
+            else:
+                varInd = self.get_variable_index(name)
+
+            dataInd = self._dataInfo[1][varInd]
```
chria, line 1562: From here on down, is this not almost the same code as in get_variable_data?

tests/test_io.py
```diff
@@ -1690,3 +1691,175 @@ def test_csv_options_cs2(self):
         simple_alias = Dummy_FMUModelCS2([("x", "y")], os.path.join(file_path, "files", "FMUs", "XML", "CS2.0", "NegatedAlias.fmu"), _connect_dll=False)
         self._run_negated_alias(self, simple_alias)
     """
+
+class TestResultDymolaBinary:
```
chria, line 1695: It seems like we only test that we can retrieve trajectories. Can we add checks so that the expected trajectories are returned?

src/common/io.py
modelonrobinandersson: I did have that initially but I removed that check with the following reason: If we want `start_index` and `stop_index` to behave like Python-slices i.e. `[start_index:stop_index]`, then we should not check that, since `array[5:0]` does not raise an error in Python for arbitrary arrays/lists.

src/common/io.py
modelonrobinandersson: This has been fixed now

tests/test_io.py
```diff
@@ -1690,3 +1691,378 @@ def test_csv_options_cs2(self):
...
+        diag_data_ratio = diag_data_ratio
+        counter = 0
+        start_index = 0
+
+        nbr_of_points = 5
+        h = 1/nbr_of_points * 0.1 # "arbitrary" step through time
+
+        data_to_return = {}
+        for i in range(nbr_of_calls):
+            time.sleep(0.1) # without this the tests fail on Linux, should look into a proper way of solving this
```
modelonrobinandersson, line 1950: I don't like this line here at all, but it is necessary in order to have the tests pass on Linux. I am open to other ideas or leave this as a TODO/follow-up ticket for now.

The reason why this is necessary is because on Linux, in the following function:
```
def _verify_file_data(self):
    if self._mtime != os.path.getmtime(self._fname): # this doesn't evaluate to True on Linux, because os.path.getmtime does not update as 'expected' (compared to Windows)
```
```

File: src/common/io.py [hunk 10]
```diff
@@ -1337,43 +1326,60 @@ def _get_name_dict(self):
...
                 return self._data_2[data_index]
 
             file_position  = self._data_2_info["file_position"]
             sizeof_type    = self._data_2_info["sizeof_type"]
             nbr_points     = self._data_2_info["nbr_points"]
             nbr_variables  = self._data_2_info["nbr_variables"]
-            
-            self._data_2[data_index] = fmi_util.read_trajectory(encode(self._fname), data_index, file_position, sizeof_type, int(nbr_points), int(nbr_variables))
+
+            # Account for sub-sets of data
```
modelonrobinandersson, line 1341: This is the "optimization" part but for regular trajectories (when diagnostics is disabled)

src/common/io.py
```diff
@@ -1382,37 +1388,48 @@ def _read_trajectory_data(self, data_index, read_diag_data):
         nbr_diag_points    = int(self._data_3_info.shape[0])
         nbr_diag_variables = int(self._data_4_info.shape[0])
 
+        # TODO account for start_index and stop_index
```
PeterMeisrimelModelon, line 1391: follow-up?

src/common/io.py
```diff
@@ -1382,37 +1388,48 @@ def _read_trajectory_data(self, data_index, read_diag_data):
...
+        data             = self._read_trajectory_data(data_index, False, start_index, stop_index)
+
         f = scipy.interpolate.interp1d(time_vector, data, fill_value="extrapolate")
-        
-        self._data_2[data_index] = f(diag_time_vector)
-        
-        return self._data_2[data_index]
 
+        self._data_2[data_index] = f(diag_time_vector)
+        return self._data_2[data_index][start_index:stop_index]
```
PeterMeisrimelModelon, line 1425: Maybe add a comment what happens here with the slicing for the default values of start_index, stop_index as None? This is not obvious

src/common/io.py
```diff
@@ -1422,14 +1439,42 @@ def _get_description(self):
...
+
+        variable_index = self.get_variable_index(variable_name)
+        data_mat = self._dataInfo[0][variable_index]
+
+        data_index = self._dataInfo[1][variable_index]
+        data_index, factor = self._correct_index_for_alias(data_index)
+
+        if data_mat == 0:
+            # Take into account that the 'Time' variable has data matrix index 0
+            # and that 'time' is called 'Time' in Dymola results
```
PeterMeisrimelModelon, line 1453: Dymola results = ResultHandler types containing "Dymola" in the name or the actual tool Dymola?

tests/test_io.py
```diff
@@ -1690,3 +1691,378 @@ def test_csv_options_cs2(self):
...
+        if only_diagnostics_data:
+            vars_to_test = ['@Diagnostics.step_time', '@Diagnostics.nbr_steps']
+        else:
+            vars_to_test = ['time', 'J4.phi']
+            if opts['dynamic_diagnostics']:
+                vars_to_test += ['@Diagnostics.step_time', '@Diagnostics.nbr_steps']
+
+        nbr_of_calls = 10
+        counter = 0
+        diag_data_ratio = np.random.randint(2, 5) # random for testing purpose
```
PeterMeisrimelModelon, line 1769: seed for consistency?

tests/test_io.py
```diff
@@ -1690,3 +1691,378 @@ def test_csv_options_cs2(self):
...
+
+                counter += 1
+                fmu.set('J4.phi', math.sin(5*fmu.time + math.pi/2)) # arbitrary
+                fmu.time += h
+            trajectories, new_start_index = rdb.get_variables_data(vars_to_test, start_index, None)
+        if update_start_index:
+            start_index = new_start_index
+
+    @testattr(stddist = True)
+    def test_get_variables_data_0(self):
```
PeterMeisrimelModelon, line 1802: I might be missing something here, but what exactly is being tested here?

The helper function just generated data, there are no asserts?

Also, I think it would be useful if the naming/docstrings would describe the practical scenario being tested, e.g., low/high frequency diagnostics, continuous fetching, etc...

tests/test_io.py
PeterMeisrimelModelon: Maybe requires a follow-up to look into an alternative to verify file data that also works on linux?

Also, shouldn't the sleep be in between the integration/diagnostics point & getting the variable data?

tests/test_io.py
```diff
@@ -1690,3 +1691,378 @@ def test_csv_options_cs2(self):
...
+
+        reference_data = {
+            0: [ 1.00000000,  0.99875026,  0.93937271],
+            1: [ 0.7960838,   0.58168309],
+            2: [ 0.31532236],
+            3: [ 0.02079483, -0.27559025],
+            4: [-0.54735767, -0.77023125]
+        }
+
+        # Just verify results for J4.phi here, but we retrieve all four trajectories at once
```
PeterMeisrimelModelon, line 1996: Do we still need the loop below then? 
The code appears more complicated than it should be here?

tests/test_io.py
```diff
@@ -1690,3 +1691,378 @@ def test_csv_options_cs2(self):
...
+            nbr_of_calls : int
+                Number of simulation calls to perform.
+
+            diag_data_ratio : int
+                Frequency of diagnostic data logging.
+
+            vars_to_test : list
+                Variables to retrieve data from during the simulation.
+
+            u : callable
```
PeterMeisrimelModelon, line 1885: Maybe use a more meaningful name? E.g., stop_index_function?

src/common/io.py
```diff
@@ -1337,43 +1326,60 @@ def _get_name_dict(self):
...
 
             file_position  = self._data_2_info["file_position"]
             sizeof_type    = self._data_2_info["sizeof_type"]
             nbr_points     = self._data_2_info["nbr_points"]
             nbr_variables  = self._data_2_info["nbr_variables"]
-            
-            self._data_2[data_index] = fmi_util.read_trajectory(encode(self._fname), data_index, file_position, sizeof_type, int(nbr_points), int(nbr_variables))
+
+            # Account for sub-sets of data
+            if start_index is not None and start_index > 0:
```
PeterMeisrimelModelon, line 1342: Also in a few other places; do we need these constant checks if start_index might be None? The default is 0?

PeterMeisrimelModelon (commented):
Looks solid! 

Various smaller comments on various naming/testing aspects.

Also, a Changelog entry is missing.

tests/test_io.py
modelonrobinandersson: Yes I am planning to create a follow-up ticket, Ill do that as the last part of the review.
That is what I thought too, but the sleep between `*_point()` and getting the data doesn't help which makes this entire thing even more odd.

src/common/io.py
modelonrobinandersson: Yup! Will be part of follow-up ticket

src/common/io.py
modelonrobinandersson: I thought about it but since we have this in 7 places right now with `[start_index:stop_index]` I stopped myself from writing that comment. Let me know what you think. I also agree it wasn't obvious to me either but it makes perfect sense after testing it.

src/common/io.py
modelonrobinandersson: I actually don't know, this comment was here in the code before I started working on it. Perhaps @chria can shed some light on this.

src/common/io.py
modelonrobinandersson: Good point, that was an old artifact. I have removed it.

tests/test_io.py
modelonrobinandersson: Good suggestion, fixed!

tests/test_io.py
modelonrobinandersson: The loop we need in order to test data from all 5 calls, since we loop over "sets of generated data".
Note also we add all of those variables to the call just to make sure the data at least for 1 of the variables (`J4.phi`) is as expected when asking for multiple variables at once.

tests/test_io.py
modelonrobinandersson: The first sets of test just verifies that we can retrieve the data, when the test scenario is constructed in different ways, i.e.:
1. No diagnostics data
2. With diagnostics data
3. More diagnostics data than regular data
4. Only diagnostics data but no regular data
There was also some randomness added here because we wanted to make sure it wasn't something as "exactly half the amount of diagnostics data as regular data".
We don't check the values in these tests since we tests that check values further down. I see why it might look a bit redundant though so let me know if you think we should just remove those tests.
When developing the tests however it was helping to make sure we don't have segmentation faults for rare cases where the ratio between regular data and diagnostics data would for some reason end up crashing together with different inputs for `start_index` and `stop_index`.

tests/test_io.py
modelonrobinandersson: Fixed

src/common/io.py
PeterMeisrimelModelon: I think it still won't be obvious to the next person stumbling upon this, including us in > 2 months; I'd suggest to just add it, even it requires to be in a few places.

tests/test_io.py
PeterMeisrimelModelon: 1. I don't fancy randomness here since it adds the risk that for (un)fortunate values a test may erroneously pass when it should not. Sure, we cannot test all combinations of integration vs. diagnostics points, but I'd still prefer having fixed and/or deterministic ratios, so we know some important cases are covered. There will always be a limitation on how many you can practically test.
2. To me it sounds like the first set of tests that only generate the data is just a single step in the second test and you kept it for historic reasons? If there was an issue in data generation, shouldn't it also occur in the second set of tests? If there is a distinction and it tests particular integration/diagnostics point ratios, then this would simply need some extra documentation.

src/common/io.py
modelonrobinandersson: What about in the docstring for `get_variables_data` where the arguments `start` and `stop` are exposed, we explain that `start` and `stop` acts like slices, i.e. `[start:stop]`, and clarify that inputs such as `[None:None]` or `[5:None]` are all valid calls? And we could even refer to https://docs.python.org/3/library/functions.html#slice where it says that `start` and `stop` default to `None`?

tests/test_io.py
modelonrobinandersson: Yeah I agree with you, randomness in tests can be nasty. The `.seed` has been added.
I have also removed the redundant tests now since I agree they are now better covered by the tests that also check the output from `get_variables_data`.

I also added CHANGELOG.

tests/test_io.py
filip-stenstrom: What is it that drives the simulation and generates new data for `result_handler.integration_point()` to write? I would assume some fmiCompletedIntegratorStep, but don't see any, but maybe it's hidden away in the dummy FMU class.  Given that you need to sleep at the start of the loop, maybe this could be related?

CHANGELOG
```diff
@@ -1,4 +1,6 @@
 --- CHANGELOG ---
+--- Future ---
+    * Added method ResultDymolaBinary.get_variables_data. Included some minor refactorization.
```
PeterMeisrimelModelon, line 3: Could maybe add a few words on what it does?

src/common/io.py
PeterMeisrimelModelon: I think in the docstrings + that start, stop act as slices with `None` being no restriction sounds good.

src/common/io.py
```diff
@@ -179,16 +180,16 @@ def get_variable_index(self,name):
                 raise VariableNotFoundError("Cannot find variable " +
                                         name + " in data file.")
 
+    def _correct_index_for_alias(self, data_index: int) -> tuple[int, int]:
+        """ Correct data_index for aliases. """
```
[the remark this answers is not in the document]
chria: The first point here is that any negative number on the data index means that it is an alias variable. The second point is that the variable it is aliased with is located in the abs position of the data index (shifted by one).

src/common/io.py
chria: Did we not remove this to a follow-up?

src/common/io.py
chria: Actual tool. But I think we can just update the comment to say that we take into account that the "time" variable can be both named "time" and "Time"

src/common/io.py
chria: You are right that it this follows the Python "standard" for slicing. However, I do not think it makes sense to make a call with stop_index < start_index and thus should result in error

src/common/io.py
modelonrobinandersson: We did, for diagnostics data. It works as expected for regular data.

src/common/io.py
modelonrobinandersson: Fixed

src/common/io.py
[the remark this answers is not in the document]
modelonrobinandersson: Thanks!

src/common/io.py
modelonrobinandersson: It was already part of the docstring but I made it even more clear now

CHANGELOG
modelonrobinandersson: Updated

PeterMeisrimelModelon (approved):

tests/test_io.py
modelonrobinandersson: In this test case we don't simulate the `normal way`, since we dont use `.simulate()`. We step manually by invoking `integration_point`, manually advance the time, update one of the state variables etc.

modelonrobinandersson: I was about to test one final thing but I can no longer reproduce the issue even when I remove the `sleep`. I wonder if this is caused by the tests that I removed in this commit https://github.com/modelon-community/PyFMI/pull/267/commits/a7524eff12b2454b70be32fc554cc9ceaee17753.

They are different tests but they are very similar, except that they dont check the returned values.

modelonrobinandersson: I see now that the tests from github actions fail, which is different from my own Ubuntu machine. I will add `sleep` back...

PR #268, opened 2024-10-11
Added option to limit the result size

chria:


tests/test_io.py
```diff
@@ -1899,3 +1899,142 @@ def test_get_variables_data_values4(self):
...
+
+            max_size = 1e6
+            opts["result_max_size"] = max_size
+            opts["ncp"] = 10000
+
+            with nose.tools.assert_raises(ResultSizeError):
+                res = model.simulate(options=opts)
+
+            result_file = model.get_last_result_file()
+            if result_file:
```
PeterMeisrimelModelon, line 1945: Not sure I see the point of the `if` here? If we expect there to be a return, we can either skip it or should have another assert for it.

tests/test_io.py
```diff
@@ -1899,3 +1899,142 @@ def test_get_variables_data_values4(self):
...
+                file_size = os.path.getsize(result_file)
+
+                assert file_size > max_size*0.9 and file_size < max_size*1.1, \
+                        "The file size is not within 10% of the given max size"
+        """
```
PeterMeisrimelModelon, line 1950:
```suggestion
        # TODO: Pytest parametrization
        """
```
Pytest parametrizations will reduce this to 2 tests + parametrization decorators, so a lot more compact.

src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -131,6 +131,12 @@ class AssimuloFMIAlgOptions(OptionBase):
             This MUST be provided.
             Default: None
 
+        result_max_size --
+            Maximum size of the stored result (in bytes). This is not a hard limit, the
+            actual size will be slightly larger to account for that the result need to
+            be consistent.
+            Default: 2GB
```
modelonrobinandersson, line 138: I think we should write:
`Default: 1024**3*2 (2GB)`

tests/test_io.py
modelonrobinandersson: I agree with Peter, I think we should force the user to have assimulo installed, since it is a dependency anyway.

src/common/io.py
```diff
@@ -2757,10 +2759,9 @@ def diagnostics_point(self, diag_data):
         """ Generates a data point for diagnostics data by invoking the util function save_diagnostics_point. """
         self.dump_data_internal.save_diagnostics_point(diag_data)
         self.nbr_diag_points += 1
-        self._make_diagnostics_consistent()
+        self._make_consistent(diag=True)
 
-    def _make_consistent(self):
-        """ TODO Requires documentation. """
+    def _make_consistent(self, diag=False):
```
modelonrobinandersson, line 2764: Can we add a docstring to what this function does? In general I think a lot of functions in PyFMI require some documentation, in order to improve understanding.

src/common/io.py
```diff
@@ -1798,13 +1803,17 @@ def integration_point(self, solver = None):
...
-        ANd this method is called after the simulation has completed.
-        """
-        pass
+        
+        try:
+            current_size = sys.getsizeof(self.time_sol) + sys.getsizeof(self.real_sol) + \
+                           sys.getsizeof(self.int_sol)  + sys.getsizeof(self.bool_sol) + \
+                           sys.getsizeof(self.param_sol)
+            
+            if current_size > self.options["result_max_size"]:
```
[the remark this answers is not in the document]
chria: Fixed

src/common/io.py
```diff
@@ -1798,13 +1803,17 @@ def integration_point(self, solver = None):
...
+                #Make the file consistent
```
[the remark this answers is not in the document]
chria: Fixed

src/common/io.py
```diff
@@ -1993,22 +1985,27 @@ def integration_point(self, solver = None):
...
-            f.write(self.const_str[:-1]+"\n")
+            self._write(self.const_str[:-1]+"\n")
         else:
-            f.write(self.const_str)
-            f.write(cont_str[:-1]+"\n")
+            self._write(self.const_str)
+            self._write(cont_str[:-1]+"\n")
 
+        try:
+            if self._current_file_size > self.options["result_max_size"]:
```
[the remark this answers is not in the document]

src/common/io.py
```diff
@@ -2770,30 +2771,23 @@ def _make_consistent(self):
...
-        t = np.array([float(self.model.time)])
-        self.dump_data(t)
-
-        f.seek(self.data_3_header_position)
-        self._data_3_header["mrows"] = self.nbr_diag_points
-        self.__write_header(self._data_3_header, "data_3")
-
-        #Reset file pointer
-        f.seek(file_pos)
+        try:
```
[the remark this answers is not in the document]

src/pyfmi/fmi_algorithm_drivers.py
chria: Updated.

src/common/io.py
chria: Added.

src/common/io.py
```diff
@@ -1798,13 +1803,17 @@ def integration_point(self, solver = None):
...
-        pass
+        
+        try:
+            current_size = sys.getsizeof(self.time_sol) + sys.getsizeof(self.real_sol) + \
+                           sys.getsizeof(self.int_sol)  + sys.getsizeof(self.bool_sol) + \
+                           sys.getsizeof(self.param_sol)
+            
+            if current_size > self.options["result_max_size"]:
+                #Make the file consistent
+                raise ResultSizeError("Maximum size of the result file reached. To change the maximum allowed result file size, please use the option 'result_max_size'")
```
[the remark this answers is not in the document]
chria: I made an update, how about now?

chria:
@PeterMeisrimelModelon I added the option for the master algorithm. For coupled simulations it should already work as it uses the same algorithm driver as a normal simulation. As for the diagnostics variables, with the refactoring done here, they will use the same code path for checking the size so there is no need to verify this separetly.

src/common/io.py
```diff
@@ -1798,13 +1803,18 @@ def integration_point(self, solver = None):
...
+        if max_size is not None:
+            current_size = sys.getsizeof(self.time_sol) + sys.getsizeof(self.real_sol) + \
+                            sys.getsizeof(self.int_sol)  + sys.getsizeof(self.bool_sol) + \
+                            sys.getsizeof(self.param_sol)
+            
+            if current_size > max_size:
+                raise ResultSizeError("Maximum size of the result reached (limit: %g GB) at time t=%g. " 
+                                  "To change the maximum allowed result size, please use the option " 
+                                  "'result_max_size' or consider reducing the number of communication "
+                                  "points alternatively the number of variables to store result for."%(max_size/1024**3, self.model.time))
```
PeterMeisrimelModelon, line 1817:
```suggestion
                raise ResultSizeError("Maximum size of the result reached (limit: %g GB) at time t=%g. " 
                                  "To change the maximum allowed result size, please use the option " 
                                  "'result_max_size', consider reducing the number of communication "
                                  "points or alternatively the number of variables to store result for."%(max_size/1024**3, self.model.time))
```

minor suggestion; Also to be applied for the other occasions.

src/common/io.py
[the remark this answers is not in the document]
PeterMeisrimelModelon: Looks great! Minor suggestion in other review comment!

PeterMeisrimelModelon (approved):

PeterMeisrimelModelon (changes requested):
Almost forgot; we'd need a changelog entry, otherwise good to merge!

src/common/io.py

PeterMeisrimelModelon (approved):
Looks great!

PR #270, opened 2024-10-22
Added check for result size at the start of a simulation

chria:


src/common/io.py
```diff
@@ -1807,13 +1813,22 @@ def integration_point(self, solver = None):
         max_size = self.options.get("result_max_size", None)
         if max_size is not None:
             current_size = sys.getsizeof(self.time_sol) + sys.getsizeof(self.real_sol) + \
-                            sys.getsizeof(self.int_sol)  + sys.getsizeof(self.bool_sol) + \
-                            sys.getsizeof(self.param_sol)
+                           sys.getsizeof(self.int_sol)  + sys.getsizeof(self.bool_sol) + \
+                           sys.getsizeof(self.param_sol)
+            if self._first_point:
```
PeterMeisrimelModelon, line 1818: Except for the "how" of determining the sizes, the code appears identical for all result handlers. I'd suggest to break it out into an independent function instead.

Additionally, I'd prefer to have the magic number of 5000 defined as e.g., "NCP_LARGE = 5000" defined somewhere in the beginning of the file.

tests/test_io.py
```diff
@@ -1931,51 +1932,77 @@ def _test_result_exception(self, result_type, result_file_name="", fmi_type="me"
...
+                    if opts["dynamic_diagnostics"]:
+                        result_handler.diagnostics_point(np.array([val[0] for val in diag_vars.values()], dtype=float))
 
             result_file = model.get_last_result_file()
-            
             file_size = os.path.getsize(result_file)
 
             assert file_size > max_size*0.9 and file_size < max_size*1.1, \
-                    "The file size is not within 10% of the given max size"
-        # TODO: Pytest parametrization
```
PeterMeisrimelModelon, line 1950: Please keep this line :)

tests/test_io.py
```diff
@@ -1931,51 +1932,77 @@ def _test_result_exception(self, result_type, result_file_name="", fmi_type="me"
             with nose.tools.assert_raises(ResultSizeError):
                 res = model.simulate(options=opts)
 
-        def _test_result_size_verification(self, result_type, result_file_name=""):
+        def _test_result_size_verification(self, result_type, result_file_name="", dynamic_diagnostics=False):
```
PeterMeisrimelModelon, line 1935: Can we maybe add a docstring what is being tested here?

tests/test_io.py
```diff
@@ -1931,51 +1932,77 @@ def _test_result_exception(self, result_type, result_file_name="", fmi_type="me"
...
+                opts['CVode_options']['rtol'] = 1e-6
+                opts['CVode_options']['atol'] = model.nominal_continuous_states * opts['CVode_options']['rtol']
+                diag_params, diag_vars = setup_diagnostics_variables(model, 0, opts, opts['CVode_options'])
+                result_handler.simulation_start(diag_params, diag_vars)
+            else:
+                result_handler.simulation_start()
 
             with nose.tools.assert_raises(ResultSizeError):
-                res = model.simulate(options=opts)
+                for i in range(ncp):
```
PeterMeisrimelModelon, line 1961: Since we now have 2 distinct `ResultSizeError` sources, I think we should check that we do in fact raise/catch the intended one when testing.

tests/test_io.py
```diff
@@ -1931,51 +1932,77 @@ def _test_result_exception(self, result_type, result_file_name="", fmi_type="me"
...
+        def _test_result_size_early_abort(self, result_type, result_file_name=""):
+            model, opts = self._setup(result_type, result_file_name)
 
             max_size = 1e6
             opts["result_max_size"] = max_size
-            opts["dynamic_diagnostics"] = True
-            opts["ncp"] = 10000
+            opts["ncp"] = 10000000
 
             with nose.tools.assert_raises(ResultSizeError):
```
PeterMeisrimelModelon, line 1980: Since we now have 2 distinct `ResultSizeError` sources, I think we should check that we do in fact raise/catch the intended one when testing.

PeterMeisrimelModelon (changes requested):
+ Changelog entry missing.

Looks good overall, a few comments for hopefully better future maintenance.

src/common/io.py

tests/test_io.py
chria: Fixed.e

tests/test_io.py
chria: Added.

tests/test_io.py
chria: This is tested in such a way that the result file is verified to be within bounds. I'm hesitant to do a check on the actual Exception message.

tests/test_io.py
chria: See above. (this is also verified by looking at the resulting file size)

tests/test_io.py
PeterMeisrimelModelon: Right, seems OK then!

PeterMeisrimelModelon (approved):

PR #272, opened 2024-10-30
Updated interface to get_variables_data

modelonrobinandersson:
Accounts for invalid range of start_index and stop_index, also returns a dict instead of a list.
Fixes issue with cached data for partial trajectories.

### DIFF
--- a/src/common/io.py
+++ b/src/common/io.py
@@ -1630,13 +1630,23 @@ class ResultDymolaBinary(ResultDymola):
         for name in names:
             trajectories[name] = self._get_variable_data_as_trajectory(name, time, start_index, stop_index)
 
-        largest_trajectory_length = -1
-        for v, t in trajectories.items():
-            largest_trajectory_length = max(largest_trajectory_length, len(t.x))
+        largest_trajectory_length = self._find_max_trajectory_length(trajectories)
         new_start_index = (start_index + largest_trajectory_length) if trajectories else start_index
 
         self._last_set_of_indices = (start_index, stop_index) # update them before we exit
         return trajectories, new_start_index
+
+    def _find_max_trajectory_length(self, trajectories):
+        """
+            Given a dict of trajectories, find the length of the largest trajectory
+            among the set of continuous variables. We disregard parameters/constants since they are not stored
+            with the same amount of data points as trajectories for continuous variables.
+        """
+        length = 0
+        for var_name, trajectory in trajectories.items():
+            if self.is_variable(var_name): # since we only consider continuous variables
+                length = max(length, len(trajectory.x))
+        return length
 
     def _calculate_events_and_steps(self, name):
         if name in self._data_3:
@@ -1712,15 +1722,13 @@ class ResultDymolaBinary(ResultDymola):
             return True
         elif '{}.'.format(DiagnosticsBase.calculated_diagnostics['nbr_state_limits_step']['name']) in name:
             return True
+
         variable_index  = self.get_variable_index(name)
         data_mat = self._dataInfo[0][variable_index]
-        if data_mat<1:
+        if data_mat < 1:
             data_mat = 1
 
-        if data_mat == 1:
-            return False
-        else:
-            return True
+        return data_mat != 1
 
     def is_negated(self, name):
         """
--- a/src/common/io.py
+++ b/src/common/io.py
@@ -1354,6 +1354,9 @@ class ResultDymolaBinary(ResultDymola):
             start_index = max(0, start_index)
             stop_index = max(0, nbr_points if stop_index is None else min(nbr_points, stop_index))
             new_file_position = file_position + start_index*sizeof_type*nbr_variables
+            # Finally when stop_index = None, we can end up with start > stop,
+            # therefore we need to use min(start, stop)
+            start_index = min(start_index, stop_index)
             new_nbr_points = stop_index - start_index
 
             self._data_2[data_index] = fmi_util.read_trajectory(
@@ -1601,7 +1604,6 @@ class ResultDymolaBinary(ResultDymola):
         if isinstance(start_index, int) and isinstance(stop_index, int) and stop_index < start_index:
             raise ValueError(f"Invalid values for {start_index=} and {stop_index=}, " + \
                               "'start_index' needs to be less than or equal to 'stop_index'.")
-
 
         trajectories = {}
 
--- a/src/common/io.py
+++ b/src/common/io.py
@@ -1332,8 +1332,8 @@ class ResultDymolaBinary(ResultDymola):
 
         return name_dict
 
-    def _can_use_partial_cache(self, start_index, stop_index):
-        """ Checks if start_index and stop_oindex are equal to the last cached indices. """
+    def _can_use_partial_cache(self, start_index: int, stop_index: Union[int, None]):
+        """ Checks if start_index and stop_index are equal to the last cached indices. """
         return self._allow_file_updates and (self._last_set_of_indices == (start_index, stop_index))
 
     def _get_trajectory(self, data_index, start_index = 0, stop_index = None):
@@ -1373,10 +1373,12 @@ class ResultDymolaBinary(ResultDymola):
         """ Returns trajectory for the diagnostics variable that corresponds to index 'data_index'. """
         self._verify_file_data()
 
-        if data_index in self._data_3:
+        index_in_cache = data_index in self._data_3
+        partial_cache_ok = self._can_use_partial_cache(start_index, stop_index)
+        if (index_in_cache and not self._allow_file_updates) or (index_in_cache and partial_cache_ok):
             return self._data_3[data_index]
-        self._data_3[data_index] = self._read_trajectory_data(data_index, True, start_index, stop_index)
-        return self._data_3[data_index][start_index:stop_index]
+        self._data_3[data_index] = self._read_trajectory_data(data_index, True, start_index, stop_index)[start_index:stop_index]
+        return self._data_3[data_index]
 
     def _read_trajectory_data(self, data_index, read_diag_data, start_index = 0, stop_index = None):
         """ Reads corresponding trajectory data for variable with index 'data_index',
@@ -1413,11 +1415,13 @@ class ResultDymolaBinary(ResultDymola):
 
         return data
 
-    def _get_interpolated_trajectory(self, data_index: int, start_index: int = None, stop_index: int = None) -> Trajectory:
+    def _get_interpolated_trajectory(self, data_index: int, start_index: int = 0, stop_index: int = None) -> Trajectory:
         """ Returns an interpolated trajectory for variable of corresponding index 'data_index'. """
         self._verify_file_data()
 
-        if data_index in self._data_2:
+        index_in_cache = data_index in self._data_2
+        partial_cache_ok = self._can_use_partial_cache(start_index, stop_index)
+        if (index_in_cache and not self._allow_file_updates) or (index_in_cache and partial_cache_ok):
             return self._data_2[data_index]
 
         diag_time_vector = self._get_diagnostics_trajectory(0, start_index, stop_index)
@@ -1426,8 +1430,9 @@ class ResultDymolaBinary(ResultDymola):
 
         f = scipy.interpolate.interp1d(time_vector, data, fill_value="extrapolate")
 
+        # note that we dont need to slice here because diag_time_vector is already sliced accordingly
         self._data_2[data_index] = f(diag_time_vector)
-        return self._data_2[data_index][start_index:stop_index]
+        return self._data_2[data_index]
 
     def _get_description(self):
         if not self._description:
--- a/src/common/io.py
+++ b/src/common/io.py
@@ -1595,33 +1595,16 @@ class ResultDymolaBinary(ResultDymola):
                               "'start_index' needs to be less than or equal to 'stop_index'.")
         trajectories = {}
 
-        # First we need to check that start_index and stop_index are within a valid range of [0, <number of data points> -1]
-        # Another way to do it is to use data_2_info and data_3_info but then we also need to invoke verify_file_data,
-        # unclear what is the most efficient approach for now.
-        if not self._contains_diagnostic_data:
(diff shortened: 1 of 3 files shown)


src/common/io.py
```diff
@@ -1344,17 +1344,10 @@ def _get_trajectory(self, data_index, start_index = 0, stop_index = None):
...
-            else:
-                new_file_position = file_position
-                new_nbr_points = nbr_points
-
-            if stop_index is not None and stop_index > 0:
-                new_nbr_points = stop_index
-                if start_index > 0:
-                    new_nbr_points -= start_index
+            start_index = max(0, start_index)
+            stop_index = max(0, nbr_points if stop_index is None else min(nbr_points, stop_index))
```
modelonrobinandersson, line 1348: One part of me thinks this should be `stop_index = max(0, nbr_points-1 if stop_index is None else min(nbr_points-1, stop_index))`, but tests only pass with implementation as above. Not sure where I am thinking incorrectly.

chria: Lets do a test. Create a test which have 100 result points. Try this method with using stop_index = 99,100,101. What do you get? What should you get?

src/common/io.py
```diff
@@ -2521,6 +2529,9 @@ class ResultSizeError(JIOError):
     Exception that is raised when a set maximum result size is exceeded.
     """
 
+class InvalidIndexError(JIOError):
```
modelonrobinandersson, line 2532: TODO remove

src/common/io.py
```diff
@@ -1607,9 +1612,10 @@ def get_variables_data(self,
             stop_index = len(time) + start_index
 
         for name in names:
-            trajectories.append(self._get_variable_data_as_trajectory(name, time, start_index, stop_index))
+            trajectories[name] = self._get_variable_data_as_trajectory(name, time, start_index, stop_index)
 
         new_start_index = start_index + len(time) if len(trajectories) > 0 else None
+        self._last_set_of_indices = (start_index, stop_index) # update them before we exit
```
chria, line 1620: Should this not take into account that we've may already have read parts of the trajectory? I.e. this should not be the union of what is already read together with this new part?

src/common/io.py
```diff
@@ -1331,11 +1332,17 @@ def _get_name_dict(self):
 
         return name_dict
 
+    def _can_use_partial_cache(self, start_index, stop_index):
+        """ Checks if start_index and stop_oindex are equal to the last cached indices. """
```
chria, line 1336:
```suggestion
        """ Checks if start_index and stop_index are equal to the last cached indices. """
```

modelonrobinandersson: Fixed

src/common/io.py
chria: And is this what you would've expect?

src/common/io.py
chria: This is what I mean. However, we've read a trajectory so it should be easy to very that it it contains what has been asked, is it not?

src/common/io.py
modelonrobinandersson: Yes

src/common/io.py
```diff
@@ -1373,10 +1373,12 @@ def _get_diagnostics_trajectory(self, data_index, start_index = 0, stop_index =
         """ Returns trajectory for the diagnostics variable that corresponds to index 'data_index'. """
         self._verify_file_data()
 
-        if data_index in self._data_3:
+        index_in_cache = data_index in self._data_3
+        partial_cache_ok = self._can_use_partial_cache(start_index, stop_index)
+        if (index_in_cache and not self._allow_file_updates) or (index_in_cache and partial_cache_ok):
```
modelonrobinandersson, line 1378: Now caching also works as expected with start/stop, including the diagnostics variables.

src/common/io.py
```diff
@@ -1607,9 +1617,10 @@ def get_variables_data(self,
             stop_index = len(time) + start_index
 
         for name in names:
-            trajectories.append(self._get_variable_data_as_trajectory(name, time, start_index, stop_index))
+            trajectories[name] = self._get_variable_data_as_trajectory(name, time, start_index, stop_index)
 
-        new_start_index = start_index + len(time) if len(trajectories) > 0 else None
+        new_start_index = stop_index if len(trajectories) > 0 else None
```
modelonrobinandersson, line 1622: The earlier formulation was also wrong, we should just use `stop_index` since we set it accordingly above.

chria: Why? What if you do not have enough points saved? I.e. there is only stop_index-3 result points in the result file? Also why you "len(traj)"? That is very confusing...

modelonrobinandersson: @chria this has been updated again, please see the latest version. I also changed it from `if len(trajectories) > 0` to `if trajectories`, since it will evaluate to `False` if `trajectories = {}`.

tests/test_io.py
```diff
@@ -1898,8 +1900,59 @@ def test_get_variables_data_values4(self):
         }
 
         for index, test_data in test_data_sets.items():
-            np.testing.assert_array_almost_equal(test_data[0], reference_data['time'][index])
-            np.testing.assert_array_almost_equal(test_data[1], reference_data['@Diagnostics.nbr_steps'][index])
+            np.testing.assert_array_almost_equal(test_data['time'].x, reference_data['time'][index])
+            np.testing.assert_array_almost_equal(test_data['@Diagnostics.nbr_steps'].x, reference_data['@Diagnostics.nbr_steps'][index])
+
+    @testattr(stddist = True)
```
modelonrobinandersson, line 1906: I added these tests as there were reports of start/stop not properly being accounted for.
However when I test this locally it works all fine for me, the reporter will come back next week after investigating if there was a user error.

src/common/io.py
```diff
@@ -1593,23 +1610,29 @@ def get_variables_data(self,
...
         else:
             # Since we interpolate data if diagnostics is enabled
             time = self._get_diagnostics_trajectory(0, start_index, stop_index)
 
-        # Need to account for data that might be added while we are iterating over 'names' later
-        if stop_index is None:
-            stop_index = len(time) + start_index
+        # If stop_index > number of data points, and data gets added while we are iterating
+        # then we might get trajectories of unequal lengths. Therefore ensure we set stop_index here accordingly.
+        stop_index = min(len(time) + start_index, float('inf') if stop_index is None else stop_index)
```
modelonrobinandersson, line 1625: @chria Added the fix we discussed. Didn't add a test, I don't see how we can test this since I need to add data when this function is being called, in fact after a particular line has been called, in this function. Let me know if you have any ideas otherwise.

src/common/io.py
```diff
@@ -1593,23 +1610,29 @@ def get_variables_data(self,
...
 
         for name in names:
-            trajectories.append(self._get_variable_data_as_trajectory(name, time, start_index, stop_index))
+            trajectories[name] = self._get_variable_data_as_trajectory(name, time, start_index, stop_index)
 
-        new_start_index = start_index + len(time) if len(trajectories) > 0 else None
+        largest_trajectory_length = -1
+        for v, t in trajectories.items():
+            largest_trajectory_length = max(largest_trajectory_length, len(t.x))
+        new_start_index = (start_index + largest_trajectory_length) if trajectories else start_index
```
modelonrobinandersson, line 1633: @chria Note the changes on line 1633

tests/test_io.py
```diff
@@ -1898,8 +1900,70 @@ def test_get_variables_data_values4(self):
...
+        # test different scenarios of start_index out of bounds
+        assert len(rdb.get_variables_data(['h'], 501, 502)[0]['h'].x) == 0
+        assert len(rdb.get_variables_data(['h'], 501, None)[0]['h'].x) == 0
+        assert len(rdb.get_variables_data(['h'], 501)[0]['h'].x) == 0
+        assert len(rdb.get_variables_data(['h'], 1234567)[0]['h'].x) == 0
+
+        # Verify next_start_index also for no variables is equal to start_index
+        assert rdb.get_variables_data([], start_index = 0)[1] == 0
+        assert rdb.get_variables_data([], start_index = 1)[1] == 1
+        assert rdb.get_variables_data([], start_index = 5)[1] == 5
```
modelonrobinandersson, line 1966: @chria The lines above verifies that `next_start_index=start_index` for `if trajectories` evalutes to False.

src/common/io.py
PeterMeisrimelModelon: Edge case question: What happens if you invoke the function for start_index = 0, stop_index = None, for a parameter and a variable, when there is only the initial data point available?

Is this a rare case where the new start index would be incorrectly adjusted for the parameter trajectory length?

modelonrobinandersson: In this case there is no result file yet. I just tested this with `fmu.initialize()`, but the result file has not been generated yet so therefore I don't think we have to worry about this edge case. Let me know if you disagree!

PeterMeisrimelModelon: I wouldn't say this is nothing to worry about, if you start simulation (after initialization) + polling for new solution points at the same time, you will very likely run into this. You'd then miss one trajectory point.

When exactly is the result file created then?

modelonrobinandersson: It looks like it is done exactly when the simulation starts.

I did another test `res = fmu.simulate(0, 0)`. Here we have `len(res['v'])` equal to 1 (bouncing ball example). Below is the output with different input:
```
# Input: start_index = 1, stop_index = 5
next_start_index=1
Len for variable='h': len(data.x)=0
Len for variable='g': len(data.x)=0
Len for variable='v': len(data.x)=0
# Input: start_index = 451, stop_index = 651
next_start_index=451
Len for variable='h': len(data.x)=0
Len for variable='g': len(data.x)=0
Len for variable='v': len(data.x)=0
# Input: start_index = 0, stop_index = None
next_start_index=1
Len for variable='h': len(data.x)=1
Len for variable='g': len(data.x)=1
Len for variable='v': len(data.x)=1
```
I would say the results are as expected, right?

src/common/io.py
chria: Great that you added it! Good with the comment. However, it is a bit confusing that you here use a "float" instead of an int. I'd recommend to change it so that we do use ints. A simple approach would be (similar to what you had before):
max_index = len(time) + start_index)
if stop_index is None:
    stop_index = max_index
else:
   stop_index = min(max_index, stop_index)

src/common/io.py
chria: I agree that this is something that we should fix. The "next index" should only consider continuous variables. As for testing, do a test instead using the specific calls to the ResultHandler routines.

PeterMeisrimelModelon: > It looks like it is done exactly when the simulation starts.
> 
> I did another test `res = fmu.simulate(0, 0)`. Here we have `len(res['v'])` equal to 1 (bouncing ball example). Below is the output with different input:
> 
> ```
> # Input: start_index = 1, stop_index = 5
> next_start_index=1
> Len for variable='h': len(data.x)=0
> Len for variable='g': len(data.x)=0
> Len for variable='v': len(data.x)=0
> # Input: start_index = 451, stop_index = 651
> next_start_index=451
> Len for variable='h': len(data.x)=0
> Len for variable='g': len(data.x)=0
> Len for variable='v': len(data.x)=0
> # Input: start_index = 0, stop_index = None
> next_start_index=1
> Len for variable='h': len(data.x)=1
> Len for variable='g': len(data.x)=1
> Len for variable='v': len(data.x)=1
> ```
> 
> I would say the results are as expected, right?

This depends on the exact call and variables.

If there is only a single data point, then `get_variables_data([<parameter>, <continuous variable>], 0, None)` is the interesting case.

src/common/io.py
modelonrobinandersson: Fixed in 81fe25e

src/common/io.py
modelonrobinandersson: Please see 51228a7

PeterMeisrimelModelon: A minor minor comments there, but looks good overall!

src/common/io.py
chria: Looking at this again - I get a bif confused. Lets say that we have 10 points in the result file. We set stop_index to None and start_index to 10. What value would stop_index get after this if-statement? Would it not get 20? (i.e. 10+10) which is more than what is in the file?

tests/test_io.py
```diff
@@ -1695,6 +1714,80 @@ def test_csv_options_cs2(self):
...
+        fmu.setup_experiment()
+        fmu.initialize()
+        opts["initialize"] = False
+
+        result_handler.set_options(opts) # required in order to call simulation_start()
+        result_handler.initialize_complete()
+        result_handler.simulation_start()
+
+        fmu.set('J4.phi', 1) # arbitrary
+        result_handler.integration_point()
```
chria, line 1744: I think that we should test getting the trajectories (using get_variables_data) before we have a call to the integration_point method.

src/common/io.py
modelonrobinandersson: No this will not be a problem since before this if-statement we have:
```
# Get the corresponding time trajectory
if not self._contains_diagnostic_data:
    time = self._get_trajectory(0, start_index, stop_index)
else:
    # Since we interpolate data if diagnostics is enabled
    time = self._get_diagnostics_trajectory(0, start_index, stop_index)

```
Here `len(time)` will be `0` since `start_index=10`. Note we also have tests for this here:
https://github.com/modelon-community/PyFMI/pull/272/files#diff-26824e205001fe146a1634cf74c1dda1cdb8a07b6216545990b773a12da705d4R2028

tests/test_io.py
modelonrobinandersson: `ResultDymolaBinary` constructor fails if you invoke it without calling `integration_point()` before. Do you have any ideas of how to get around this? We currently get an uncaught exception pointing to issues when setting up `self._data_2`.

src/common/io.py
chria: I see, thank you.

tests/test_io.py
chria: Ok, I think we should fix that. But we can leave that for a follow-up ticket.

PR #276, opened 2024-11-29
Added verification towards disk space

chria:


src/common/io.py
```diff
@@ -42,6 +43,15 @@
 
 SYS_LITTLE_ENDIAN = sys.byteorder == 'little'
 NCP_LARGE = 5000
+DISK_PROTECTION = 50*1024**2 #50mb
+
+#Get the available disk space at a given local path (minus 50mb to make sure we do not run out of space)
+def get_available_disk_space(local_path):
```
PeterMeisrimelModelon, line 49:
```suggestion
def get_available_disk_space(local_path):
    """Get the available disk space at a given local path (minus 50mb to make sure we do not run out of space)."""
```
Might as well make this a docstring.

src/common/io.py
```diff
@@ -42,6 +43,15 @@
...
+    try:
+        free_space = disk_usage(local_path).free - DISK_PROTECTION
```
PeterMeisrimelModelon, line 51: Not sure I understand what the rational behind `DISK_PROTECTION` is. 
Can we add some comment here?

src/common/io.py
```diff
@@ -42,6 +43,15 @@
...
+    except (FileNotFoundError, TypeError):
```
PeterMeisrimelModelon, line 52: Can we add some comments here on what these are meant to catch? This is not evident to me, even after checking the documentation of `disk_usage`.

PeterMeisrimelModelon (changes requested):

src/common/io.py
chria: Added doc.

src/common/io.py
chria: Added doc.

PeterMeisrimelModelon (approved):
Thanks for adding documentation! 

Can you also add a changelog entry? Good to merge then!

PR #280, opened 2025-01-10
Increased robustness of getting supports attribute from ResultHandlers

PeterMeisrimelModelon:


tests/test_io.py
```diff
@@ -2156,3 +2158,43 @@ def test_large_size_memory(self):
 
     def test_large_size_memory_stream(self):
         self._test_result("memory", StringIO())
+
+
+class ResultHandlerCustomNoSupport(ResultHandler):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.supports = {} # Don't do that
```
chria, line 2166:
```suggestion
        self.supports = {} # Don't do that (the dictionary should be UPDATED, not REPLACED)
```

PeterMeisrimelModelon:
> Looks good! Maybe though add a comment as well in the base class for the ResultHandler that the supports dictionary should not be overwritten, only updated.

Added.

PR #282, opened 2025-01-20
Fixed Exception when using dynamic_diagnostics and scalar atol

PeterMeisrimelModelon:
I considered that the conversion of scalar atol to list could also be done in the `_set_absolute_tolerance_options` function (in `fmi_algorithm_drivers.py`). 

I do however prefer this variant since the adjustment is close to where it is needed. The other variant would create an unnecessary convoluted dependency in the code.

### DIFF
--- a/src/common/diagnostics.py
+++ b/src/common/diagnostics.py
@@ -70,7 +70,11 @@ def setup_diagnostics_variables(model, start_time, options, solver_options):
                 atol = [atol]*len(states_list)
             # atol is "pseudoscalar", array/list with single entry; 
             if np.size(atol) == 1:
-                atol = [atol[0]]*len(states_list)
+                # distinction here is need since e.g., np.array(1.) does not support indexing
+                if isinstance(atol, np.ndarray):
+                    atol = [atol.item()]*len(states_list)
+                else: # general iterable, e.g., list
+                    atol = [atol[0]]*len(states_list)
                 
             _diagnostics_params[f"{DIAGNOSTICS_PREFIX}solver.relative_tolerance"] = (rtol, "Relative solver tolerance.")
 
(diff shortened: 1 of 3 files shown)


src/common/diagnostics.py
```diff
@@ -62,6 +63,8 @@ def setup_diagnostics_variables(model, start_time, options, solver_options):
             atol = solver_options.get('atol', None)
             if (rtol is None) or (atol is None):
                 rtol, atol = model.get_tolerances()
+            if isinstance(atol, numbers.Number): # is atol is scalar, convert to list
```
chria, line 66: Do we want to handle cases where atol is array(1e-4) and array([1e-4] ?

PeterMeisrimelModelon: I think we do, since we currently allow such `atol` inputs + the error messages for using a size-mismatched atol point to the fact that `atol = [1e-4]` is allowed.

`Failed to set the solver option 'atol' with msg: atol must be of length one or same as the dimension of the problem.`

Fixed; see latest commit.

tests/test_fmi.py
```diff
@@ -1117,6 +1117,17 @@ def test_atol_auto_update6(self):
         model.simulate(options=opts, algorithm=NoSolveAlg)
         np.testing.assert_allclose(opts["CVode_options"]["atol"], [0.03, 0.03])
 
+    @pytest.mark.parametrize("atol", [1e-4, [1e-4], np.array([1e-4])])
```
chria, line 1120: Maybe also np.array(1e-4) for completeness?

PeterMeisrimelModelon: Added, see latest commit. Required some code change since 0-dim numpy arrays are a surprising mess.

PR #288, opened 2025-01-28
Add test utility to retrieve reference FMUs for testing

modelonrobinandersson:


tests/test_fmi3.py
```diff
@@ -0,0 +1,13 @@
...
+
+this_dir = Path(__file__).parent.absolute()
+sys.path.append(str(this_dir))
+
+from utils import setup_reference_fmus
+
+setup_reference_fmus()
+expected_fmu = Path(this_dir) / 'files' / 'reference_fmus' / '3.0' / 'VanDerPol.fmu'
+if not expected_fmu.exists():
+    raise Exception(f"Test setup failed, FMU {expected_fmu} does not exist!")
```
modelonrobinandersson, line 13: This is just a temporary sanity check to verify it works. Once we start adding tests that involve these FMUs I think we should just remove above code.

tests/utils.py
```diff
@@ -0,0 +1,49 @@
+import requests
+import hashlib
+from tempfile import TemporaryDirectory
+from zipfile import ZipFile
+from pathlib import Path
+
+files_directory = Path(__file__).parent / 'files'
+
+def setup_reference_fmus():
```
PeterMeisrimelModelon, line 9: I think we should rename this file to `conftest.py` and give this setup function the decorator 
`@pytest.fixture(autouse="True", session = "scope")`

Then it will be automatically run exactly ONCE with every `pytest` session.

See also: https://docs.pytest.org/en/6.2.x/fixture.html?highlight=conftest%20py#scope-sharing-fixtures-across-classes-modules-packages-or-session

Please have a look and see if you think this would be a good idea.

tests/test_fmi3.py
PeterMeisrimelModelon: I think it is OK to keep, to have a simple sanity check that the download script doesn't break. 

The actual testing should be moved into a test function, so the result actually shows up with pytest.

tests/utils.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+        This function downloads reference FMUs from the Modelica group and unpacks
+        them into the test files directory.
+        The MD5sum of the URL is checked in order to avoid unnecessary downloading.
+        Note that this requires an internet connection to work.
+    """
+
+    def download_url(url, save_file_to, chunk_size=1024):
+        """ Download file from URL to 'save_file_to' in chunks. """
+        r = requests.get(url, stream=True)
```
PeterMeisrimelModelon, line 19: What happens if this fails, e.g., files were removed or no internet connection? 

I suppose having an assert on the return status (if there is one) would be OK here.

tests/test_fmi3.py
modelonrobinandersson: I have moved it now, it was also required by pytest once I added the fixture.

tests/utils.py
modelonrobinandersson: I think it would be a great idea, thanks for sharing! I am not sure if we want `autouse=True`, it says:
_Sometimes you may want to have a fixture (or even several) that you know all your tests will depend on. “Autouse” fixtures are a convenient way to make all tests automatically request them_

I see the purpose of it but I think if we simply have the proposed fixture, without autouse,  on the setup-function, we should be good? Let me know what you think I don't have any strong opinions here.

tests/utils.py
modelonrobinandersson: I have replaced the use of `requests` with `urllib` instead (part of Python standard library).
There is now also some exception handling that I have tested by turning off my internet connection.

tests/test_fmi3.py
```diff
@@ -0,0 +1,13 @@
+# TODO: Remove/rename this file
+import sys
+from pathlib import Path
+
+this_dir = Path(__file__).parent.absolute()
+sys.path.append(str(this_dir))
+
+from conftest import setup_reference_fmus
```
PeterMeisrimelModelon, line 8: This line can actually be skipped, all fixtures in conftest.py are automatically imported.

tests/test_fmi3.py
```diff
@@ -0,0 +1,13 @@
...
+
+def test_foo(setup_reference_fmus):
+    expected_fmu = Path(this_dir) / 'files' / 'reference_fmus' / '3.0' / 'VanDerPol.fmu'
+    if not expected_fmu.exists():
+        raise Exception(f"Test setup failed, FMU {expected_fmu} does not exist!")
```
PeterMeisrimelModelon, line 13:
```suggestion
    assert expected_fmu.exists()
```

tests/utils.py
PeterMeisrimelModelon: I see this creating the following issue:

One needs to use this decorator in every test that uses reference FMUs, which will likely be a few.

"autouse" gets around that. Having the decorator included on a file basis would be sensible I suppose, but we will likely be using the FMI3 reference FMUs in all FMI3 related tests sooner or later anyways.

modelonrobinandersson: I see what you're saying. I have added the "autouse" now. I think also we are fine since we no longer download the zip if not necessary, so it should be minimal overhead.

tests/test_fmi3.py
modelonrobinandersson: I removed most stuff above now, see my latest commit

tests/test_fmi3.py
modelonrobinandersson: Fixed as well

PeterMeisrimelModelon (approved):

PR #290, opened 2025-01-31
Restructure to separate code of different FMI versions (part 1/x)

PeterMeisrimelModelon:


src/pyfmi/exceptions.py
```diff
@@ -0,0 +1,63 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+# This file contains the various exceptions classes used in PyFMI
+
+class FMUException(Exception):
```
PeterMeisrimelModelon, line 20: Simply copied all Exceptions previously in fmi.pyx

src/pyfmi/fmi.pyx
```diff
@@ -27,8710 +27,89 @@ For profiling:
...
-                            elif name in inputs_list:
-                                inputs[derivatives[i]].append(name)
-                                inputs_kind[derivatives[i]].append(factor_kindp[start_indexp[i]+j])
-
-                #Free the variable list
-                FMIL.fmi2_import_free_variable_list(variable_list)
-
-        #Caching
-        self._derivatives_states_dependencies = states
-        self._derivatives_states_dependencies_kind = states
```
PeterMeisrimelModelon, line 47: Essentially all the imports for backward compatibility.

src/pyfmi/fmi2.pyx
```diff
@@ -0,0 +1,5256 @@
...
+# Evaluate Jacobian w.r.t. states.
+FMI_STATES = 1
+# Evaluate Jacobian w.r.t. inputs.
+FMI_INPUTS = 2
+# Evaluate Jacobian of derivatives.
+FMI_DERIVATIVES = 1
+# Evaluate Jacobian of outputs.
+FMI_OUTPUTS = 2
+
+# TODO: Copied from utils temporarily; to avoid circular dependencies
```
PeterMeisrimelModelon, line 110: fmi_util.pyx has some unpleasant circular dependencies with the fmi*.pyx files, this is the only place this function is used though.

Copied it here for now to get rid of the circular dependency.

src/pyfmi/fmi_base.pxd
```diff
@@ -0,0 +1,62 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+# Module containing the abstract base class for FMI interface Python wrappers.
+# Plus some auxiliary functions
+
+cimport pyfmi.fmil_import as FMIL
+
+# TODO; temporarily here due to circular dependency
```
PeterMeisrimelModelon, line 23: Same thing here, unpleasant circular dependencies with fmi_util.pyx

src/pyfmi/fmi_base.pyx
```diff
@@ -0,0 +1,788 @@
...
+cdef FMIL.fmi_version_enu_t import_and_get_version(FMIL.fmi_import_context_t* context, char* fmu_full_path, char* fmu_temp_dir, int allow_unzipped_fmu):
+    """ Invokes the necessary FMIL functions to retrieve FMI version while accounting
+        for the conditional 'allow_unzipped_fmu'.
+    """
+    if allow_unzipped_fmu:
+        return FMIL.fmi_import_get_fmi_version(context, NULL, fmu_temp_dir)
+    else:
+        return FMIL.fmi_import_get_fmi_version(context, fmu_full_path, fmu_temp_dir)
+
+# TODO: copied due to circular dependencies; replace by functools lru_cache?
```
PeterMeisrimelModelon, line 76: XXX: Could potentially be replaced, not sure how important the storing in binary from "marshal" is.

src/pyfmi/fmi_base.pyx
```diff
@@ -0,0 +1,788 @@
...
+        Prints the log information to the prompt.
+        """
+        cdef int N
+        log = self.get_log(start_lines, end_lines)
+        N = len(log)
+
+        for i in range(N):
+            print(log[i])
+    
+    def _get_module_name(self):
```
PeterMeisrimelModelon, line 425: This one is new, see other comment for usage.

src/pyfmi/fmi_base.pyx
```diff
@@ -0,0 +1,788 @@
...
+                file_path -- path to extracted XML file
+            otherwise function returns nothing
+        """
+        # XXX: Needs to be here for now, due to circular dependency, since pyfmi.common.log.parser imports FMUException (from this file)
+        from pyfmi.common.log import extract_xml_log
+
+        if file_name is None:
+            file_name = "{}.{}".format(os.path.splitext(self.get_log_filename())[0], 'xml')
+
+        module_name = self._get_module_name()
```
PeterMeisrimelModelon, line 452: was previously 

`model_name = "Model"`, unless `instance(self, FMUModelCS1)`

Replaced by this to get rid of the circular dependency with fmi1.pyx (`FMUModelCS1` overwrites the function)

src/pyfmi/fmi_util.pyx
```diff
@@ -237,10 +241,10 @@ cpdef prepare_data_info(np.ndarray[int, ndim=2] data_info, list sorted_vars, lis
     cdef int nof_diag_params = len(diagnostics_param_values)
     cdef int i, alias, data_type, variability
     cdef int last_data_matrix = -1, last_index = -1
-    cdef int FMI_NEGATED_ALIAS = fmi.FMI_NEGATED_ALIAS
-    cdef int FMI_PARAMETER = fmi.FMI_PARAMETER, FMI_CONSTANT = fmi.FMI_CONSTANT
-    cdef int FMI_REAL = fmi.FMI_REAL, FMI_INTEGER = fmi.FMI_INTEGER
-    cdef int FMI_ENUMERATION = fmi.FMI_ENUMERATION, FMI_BOOLEAN = fmi.FMI_BOOLEAN
+    cdef int _FMI_NEGATED_ALIAS = FMI_NEGATED_ALIAS # TODO
```
PeterMeisrimelModelon, line 244: These are also used in FMI2 contexts, not the only place either.

tests/test_fmi_imports.py
```diff
@@ -0,0 +1,110 @@
+import pytest
+import pyfmi.fmi as fmi
+
+@pytest.mark.parametrize("x",
```
PeterMeisrimelModelon, line 4: Will remove, but here for visibility; these are all the imports one could get from `pyfmi.fmi`, i.e., nothing broke.

src/pyfmi/fmi.pyx
modelonrobinandersson: Can we put each import on a new row? For example
```
from a import (
    b,
    c,
    ...
)
```

src/pyfmi/fmi_base.pyx
```diff
@@ -0,0 +1,787 @@
...
+            # Returns a list of the variables
+            model.get(['damper.d','gear.a'])
+        """
+        if isinstance(variable_name, basestring):
+            return self._get(variable_name) #Scalar case
+        else:
+            ret = []
+            for i in xrange(len(variable_name)): #A list of variables
+                ret += [self._get(variable_name[i])]
+            return ret
```
modelonrobinandersson, line 169: Unrelated to your MR but caught it while scrolling
```suggestion
            return [self._get(variable_name[i]) for i in xrange(len(variable_name))]
```

modelonrobinandersson (approved):
I think this looks good, and it is a great start. I agree there are more things to do but we could do those in separate MRs to simplify the review (my browser gets very slow when reviewing this).
I left a couple of minor comments, I also noticed we have a lot of comments that are like `#foo` instead `# foo` which we should try to fix but we can do that in a separate MR as well.

Well done!

src/pyfmi/fmi.pyx
PeterMeisrimelModelon: Fixed.

modelonrobinandersson (approved):

PR #295, opened 2025-02-19
Added support for initializing and terminating ME FMUs

modelonrobinandersson:
- [x] Change log ? I am also open to wait with this until more FMI3 functionality is in place.
- [x] FMIL external needs to be updated since this only works with FMIL master
- [x] Implement `terminate` function, reviewer can still review the rest.

### DIFF
--- a/src/pyfmi/fmi2.pyx
+++ b/src/pyfmi/fmi2.pyx
@@ -40,7 +40,7 @@ from pyfmi.fmi_base import (
 from pyfmi.exceptions import (
     FMUException,
     InvalidBinaryException,
-    InvalidXMLException, 
+    InvalidXMLException,
     InvalidVersionException
 )
 from pyfmi.common.core import create_temp_dir
@@ -1185,7 +1185,6 @@ cdef class FMUModelBase2(FMI_BASE.ModelBase):
 
         self._t = start_time
         self._last_accepted_time = start_time
-        self._relative_tolerance = tolerance
 
         self._log_handler.capi_start_callback(self._max_log_size_msg_sent, self._current_log_size)
         status = FMIL2.fmi2_import_setup_experiment(self._fmu,
@@ -5172,12 +5171,12 @@ cdef class WorkerClass2:
         return ret
 
 cdef object _load_fmi2_fmu(
-    fmu, 
-    object log_file_name, 
-    str kind, 
-    int log_level, 
+    fmu,
+    object log_file_name,
+    str kind,
+    int log_level,
     int allow_unzipped_fmu,
-    FMIL.fmi_import_context_t* context, 
+    FMIL.fmi_import_context_t* context,
     bytes fmu_temp_dir,
     FMIL.jm_callbacks callbacks,
     list log_data
--- a/src/pyfmi/fmi2.pyx
+++ b/src/pyfmi/fmi2.pyx
@@ -1205,7 +1205,7 @@ cdef class FMUModelBase2(FMI_BASE.ModelBase):
         status = FMIL2.fmi2_import_reset(self._fmu)
         self._log_handler.capi_end_callback(self._max_log_size_msg_sent, self._current_log_size)
         if status != 0:
-            raise FMUException('An error occured when reseting the model, see the log for possible more information')
+            raise FMUException('An error occured when resetting the model, see the log for possible more information')
 
         #Default values
         self._t = None
--- a/src/pyfmi/fmi3.pyx
+++ b/src/pyfmi/fmi3.pyx
@@ -130,7 +130,7 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
         self._context = FMIL.fmi_import_allocate_context(&self.callbacks)
         self._allocated_context = 1
 
-        #Get the FMI version of the provided model
+        # Get the FMI version of the provided model
         if _unzipped_dir:
             fmu_temp_dir = pyfmi_util.encode(_unzipped_dir)
         elif self._allow_unzipped_fmu:
@@ -174,6 +174,7 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
 
         self._fmu_kind = FMIL3.fmi3_import_get_fmu_kind(self._fmu)
         self._allocated_xml = 1
+
         # FMU kind is unknown
         if self._fmu_kind & FMIL3.fmi3_fmu_kind_unknown:
             last_error = pyfmi_util.decode(FMIL.jm_get_last_error(&self.callbacks))
@@ -184,7 +185,7 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
         else:
             self._fmu_kind = self._get_fmu_kind()
 
-        #Connect the DLL
+        # Connect the DLL
         if _connect_dll:
             self._log_handler.capi_start_callback(self._max_log_size_msg_sent, self._current_log_size)
             status = FMIL3.fmi3_import_create_dllfmu(self._fmu, self._fmu_kind, NULL, FMIL3.fmi3_log_forwarding)
@@ -231,7 +232,7 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
             self._fmu_log_name = <char*>FMIL.malloc((FMIL.strlen(fmu_log_name)+1)*sizeof(char))
             FMIL.strcpy(self._fmu_log_name, fmu_log_name)
 
-            #Create the log file
+            # Create the log file
             with open(self._fmu_log_name,'w') as file:
                 for i in range(len(self._log)):
                     file.write("FMIL: module = %s, log level = %d: %s\n" % (
--- a/src/pyfmi/fmi3.pyx
+++ b/src/pyfmi/fmi3.pyx
@@ -231,10 +231,17 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
         self._log = []
 
     def __dealloc__(self):
-        """
-        Deallocate memory
-        """
+        """ Deallocate allocated memory. """
         self._invoked_dealloc = 1
+
+        if self._initialized_fmu == 1:
+            FMIL3.fmi3_import_terminate(self._fmu)
+
+        if self._allocated_fmu == 1:
+            FMIL3.fmi3_import_free_instance(self._fmu)
+
+        if self._allocated_dll == 1:
+            FMIL3.fmi3_import_destroy_dllfmu(self._fmu)
 
         if self._allocated_xml == 1:
             FMIL3.fmi3_import_free(self._fmu)
@@ -255,17 +262,41 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
         if self._log_stream:
             self._log_stream = None
 
+    def terminate(self):
+        """
+        Calls the FMI function fmi3Terminate() on the FMU.
+        After this call, any call to a function changing the state of the FMU will fail.
+        """
+        cdef FMIL3.fmi3_status_t status
+
+        self._log_handler.capi_start_callback(self._max_log_size_msg_sent, self._current_log_size)
+        status = FMIL3.fmi3_import_terminate(self._fmu)
+        self._log_handler.capi_end_callback(self._max_log_size_msg_sent, self._current_log_size)
+
+        if status != 0:
+            raise FMUException("Termination of FMU failed, see log for possible more information.")
+
+    def free_instance(self):
+        """
+        Calls the FMI function fmi3FreeInstance() on the FMU. Note that this is not
+        needed generally.
+        """
+        self._log_handler.capi_start_callback(self._max_log_size_msg_sent, self._current_log_size)
+        FMIL3.fmi3_import_free_instance(self._fmu)
+        self._log_handler.capi_end_callback(self._max_log_size_msg_sent, self._current_log_size)
+
+
     def reset(self):
         """ Resets the FMU back to its original state. Note that the environment
-        has to initialize the FMU again after this function-call. """
+        has to initialize the FMU again after this function-call.
+        """
         cdef FMIL3.fmi3_status_t status
 
         self._log_handler.capi_start_callback(self._max_log_size_msg_sent, self._current_log_size)
         status = FMIL3.fmi3_import_reset(self._fmu)
         self._log_handler.capi_end_callback(self._max_log_size_msg_sent, self._current_log_size)
         if status != 0:
-            raise FMUException('An error occured when reseting the model, see the log for possible more information')
-
+            raise FMUException('An error occured when resetting the model, see the log for possible more information')
 
         #Default values
         self._t = None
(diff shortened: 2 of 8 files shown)


src/pyfmi/fmi2.pyx
```diff
@@ -1185,7 +1185,6 @@ cdef class FMUModelBase2(FMI_BASE.ModelBase):
 
         self._t = start_time
         self._last_accepted_time = start_time
-        self._relative_tolerance = tolerance
```
modelonrobinandersson, line 1188: I removed this as it doesnt seem used anywhere else besides just getting a value assigned in setup_experiment.

src/pyfmi/fmi3.pyx
```diff
@@ -180,12 +184,48 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
         else:
             self._fmu_kind = self._get_fmu_kind()
 
-        self._modelId = "TODO" # TODO
+        #Connect the DLL
```
modelonrobinandersson, line 187: For the reviewer: Basically the same/similar structure as in FMI2

src/pyfmi/fmi3.pyx
```diff
@@ -223,9 +266,27 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
         if self._log_stream:
             self._log_stream = None
 
+    def reset(self):
+        """ TODO """
+        self._t = None
```
modelonrobinandersson, line 271: I left the reset function like this since it is to be implemented later. But I dont want us to forget about resetting `self._t` which I had to introduce in this PR.

tests/test_fmi3.py
```diff
@@ -81,6 +90,49 @@ def test_load_kind_SE(self, ref_fmu):
...
+        """Test that model identifier is retrieved as expected."""
+        fmu = load_fmu(FMI3_REF_FMU_PATH / "VanDerPol.fmu") # any FMI3 ME would suffice
+        assert fmu.get_identifier() == 'VanDerPol'
+
+    def test_get_get_version(self):
+        """Test that FMI version is retrieved as expected."""
+        fmu = load_fmu(FMI3_REF_FMU_PATH / "VanDerPol.fmu") # any FMI3 ME would suffice
+        assert fmu.get_version() == '3.0'
+
+    def test_instantiation(self, tmpdir):
```
modelonrobinandersson, line 103: Thinking of removing this test since we have the test with `initialize` below.

src/pyfmi/fmi3.pyx
```diff
@@ -287,6 +375,189 @@ cdef class FMUModelME3(FMUModelBase3):
...
+        """
+        cdef FMIL3.fmi3_status_t status
+        self._log_handler.capi_start_callback(self._max_log_size_msg_sent, self._current_log_size)
+        status = FMIL3.fmi3_import_exit_initialization_mode(self._fmu)
+        self._log_handler.capi_end_callback(self._max_log_size_msg_sent, self._current_log_size)
+
+        if status != FMIL3.fmi3_status_ok:
+            raise FMUException("Failed to exit initialization mode")
+
+    def enter_continuous_time_mode(self):
```
modelonrobinandersson, line 541: This function and `enter_event_mode` I added before I realized I did not need them.

src/pyfmi/fmil3_import.pxd
```diff
@@ -45,23 +60,55 @@ cdef extern from 'fmilib.h':
...
+        fmi3_import_t* fmu,
+        fmi3_boolean_t toleranceDefined,
+        fmi3_float64_t tolerance,
+        fmi3_float64_t startTime,
+        fmi3_boolean_t stopTimeDefined,
+        fmi3_float64_t stopTime)
+    fmi3_status_t fmi3_import_exit_initialization_mode(fmi3_import_t* fmu)
+    fmi3_status_t fmi3_import_enter_event_mode(fmi3_import_t* fmu)
+    fmi3_status_t fmi3_import_enter_continuous_time_mode(fmi3_import_t* fmu)
+    fmi3_status_t fmi3_import_enter_event_mode(fmi3_import_t* fmu)
```
PeterMeisrimelModelon, line 95:
```suggestion
```
See 2 lines above, you seem to have added it twice by accident.

tests/test_fmi3.py
```diff
@@ -81,6 +90,49 @@ def test_load_kind_SE(self, ref_fmu):
...
+    @pytest.mark.parametrize("ref_fmu", [
+        FMI3_REF_FMU_PATH / "BouncingBall.fmu",
+        FMI3_REF_FMU_PATH / "Dahlquist.fmu",
+        FMI3_REF_FMU_PATH / "Resource.fmu",
+        FMI3_REF_FMU_PATH / "StateSpace.fmu",
+        FMI3_REF_FMU_PATH / "Feedthrough.fmu",
+        FMI3_REF_FMU_PATH / "Stair.fmu",
+        FMI3_REF_FMU_PATH / "VanDerPol.fmu",
+    ])
+    def test_load_kind_auto(self, ref_fmu):
```
PeterMeisrimelModelon, line 132:
```suggestion
    def test_initialize(self, ref_fmu):
```
Or similar

tests/test_fmi3.py
```diff
@@ -81,6 +90,49 @@ def test_load_kind_SE(self, ref_fmu):
...
+        """Test initialize all the ME reference FMUs. """
+        fmu = load_fmu(ref_fmu)
+        fmu.initialize() # Should simply pass without any exceptions
```
PeterMeisrimelModelon, line 136: I propose to add a few more unit tests for the various functions added:

1. get_default_experiment_(start|stop|tolerace) (possibly in test_fmi.py & parameterize for FMI1,2,3)
2. manually calling enter/exit init mode
3. enter_continuous_time_mode
4. enter_event_mode

PeterMeisrimelModelon (commented):
Looks great thus far. Aside from the comments a few things we should also add here:

- explicit free_instance function + accompanied appropriate test.
- fmi3_import_destroy_dllfmu & fmi3_import_free_instance call in `__dealloc__`. See FMI2 for the various occurrences of `_allocated_dll` and `_allocated_fmu`.
- [Optional] Related to `terminate`, some tests for valid/invalid call sequences, e.g., double termiante is not valid, and the FMU will only "midly" complain, not sure if this should be an error on the PyFMI side though

src/pyfmi/fmi3.pyx
```diff
@@ -180,21 +185,60 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+            if status == FMIL.jm_status_error:
+                last_error = pyfmi_util.decode(FMIL3.fmi3_import_get_last_error(self._fmu))
+                if self._enable_logging:
+                    raise InvalidBinaryException("The FMU could not be loaded. Error loading the binary. " + last_error)
+                else:
+                    raise InvalidBinaryException("The FMU could not be loaded. Error loading the binary. Enable logging for possibly more information.")
+            self._allocated_dll = 1
+
+        # Note that below, values are retrieved from XML (via FMIL) if .dll/.so is not connected
+        if self._fmu_kind & FMIL3.fmi3_fmu_kind_me:
```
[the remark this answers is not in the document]
modelonrobinandersson: That implies that we still call an internal function (instead of an internal attribute) and not sure if that looks so much better? We have a few lines up:
```
self._fmu_kind = self._get_fmu_kind()
```
Let me know what you think.

src/pyfmi/fmi3.pyx
modelonrobinandersson: Fixed

src/pyfmi/fmi3.pyx
```diff
@@ -180,21 +185,60 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+            self._modelId= pyfmi_util.decode(FMIL3.fmi3_import_get_model_identifier_ME(self._fmu))
+        else:
+            raise NotImplementedError(f"FMUModelBase3 only supports FMU type 'Model Exchange'")
+
+        self._modelName = pyfmi_util.decode(FMIL3.fmi3_import_get_model_name(self._fmu))
 
+        self._log_handler.capi_start_callback(self._max_log_size_msg_sent, self._current_log_size)
+        status = FMIL3.fmi3_import_get_number_of_event_indicators(self._fmu, &self._nEventIndicators)
```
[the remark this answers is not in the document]
modelonrobinandersson: Good point. Fixed in 2b3b66e

src/pyfmi/fmi3.pyx
```diff
@@ -180,21 +185,60 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+        if status != FMIL3.fmi3_status_ok:
+            raise InvalidFMUException("The FMU could not be instantiated, error retrieving number of event indicators.")
+
+        self._log_handler.capi_start_callback(self._max_log_size_msg_sent, self._current_log_size)
+        status = FMIL3.fmi3_import_get_number_of_continuous_states(self._fmu, &self._nContinuousStates)
+        self._log_handler.capi_end_callback(self._max_log_size_msg_sent, self._current_log_size)
+        if status != FMIL3.fmi3_status_ok:
+            raise InvalidFMUException("The FMU could not be instantiated, error retrieving number of continuous states.")
+
+        # TODO: The code below is identical between FMUModelBase2 and FMUModelBase3, perhaps we can refactor this
```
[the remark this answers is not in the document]
modelonrobinandersson: I agree, let's leave it for now I dont think its a big priority.

src/pyfmi/fmi3.pyx
```diff
@@ -287,6 +376,189 @@ cdef class FMUModelME3(FMUModelBase3):
...
+        self,
+        tolerance_defined=True,
+        tolerance="Default",
+        start_time="Default",
+        stop_time_defined=False,
+        stop_time="Default"
+    ):
+        """
+        Initializes the model and computes initial values for all variables.
+        Additionally calls the setup experiment, if not already called.
```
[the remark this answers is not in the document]
modelonrobinandersson: Removed, accidental copy pasta. Also I cannot reply to the comment above for some reason. But now are fixed in 2b3b66e.

tests/test_fmi3.py
```diff
@@ -81,6 +90,49 @@ def test_load_kind_SE(self, ref_fmu):
...
+        found_substring = False
+        substring_to_find = 'Successfully loaded all the interface functions'
+
+        with temp_dir_context(tmpdir) as temp_path:
+             # any FMI3 ME would suffice, log_level set to 5 required by test
+            fmu = load_fmu(FMI3_REF_FMU_PATH / "VanDerPol.fmu", log_level=5)
+            with open(fmu.get_log_filename(), 'r') as f:
+                contents = f.readlines()
+
+        for line in contents:
```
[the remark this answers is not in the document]
modelonrobinandersson: That looks much better indeed. Fixed in 2b3b66e.

src/pyfmi/fmil3_import.pxd
modelonrobinandersson: Fixed in 2b3b66e

tests/test_fmi3.py
modelonrobinandersson: Fixed in 2b3b66e

tests/test_fmi3.py
modelonrobinandersson: I added test for this in 2b3b66e but only for FMI 3. I don't think we should over-test these things right now as we do have tests for it but in other places.
Also I cannot add tests yet for `enter_continuous_time_mode` nor `enter_event_mode` since it is part of the next effort for the simulation (I currently get an error invalid sequence of function calls).

src/pyfmi/fmi3.pyx
modelonrobinandersson: @PeterMeisrimelModelon I added more functionality in 90e950a.

tests/test_fmi3.py
PeterMeisrimelModelon: I'd say keep it, good to have tests in smaller units.

tests/test_fmi3.py
PeterMeisrimelModelon: What is wrong with entering contnuous time more/event mode? Unless there was some change for the FMI3 state machine I missed, these should simply work when done after initialization.

src/pyfmi/fmi3.pyx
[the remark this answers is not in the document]
PeterMeisrimelModelon: Ah, my main intention with this was that any child-class specific logic should happen in that specific child. Otherwise we are simply creating circular dependencies.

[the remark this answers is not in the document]
modelonrobinandersson: Oh, yes I agree. However I cannot see an easy way to change this because `self._modelID` is used in the constructor of `FMUModelBase` in order to create the log file with appropriate name.

A solution I see would be to forward it to the constructor of `FMUModelBase` from `FMUModelME3`, if you have any other suggestions my ears are all yours. Could chat about it next week.

tests/test_fmi3.py
modelonrobinandersson: I must have done something wrong when I first tested it, I tried just now again and it works as expected, see 79f032b.

src/pyfmi/fmi3.pyx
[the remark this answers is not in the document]
PeterMeisrimelModelon: It doesn't need to be part of the class specific constructor, it can be a separate function, see e.g., `_get_fmu_kind`. Also note that in the abstract base this function directly raises an `Exception`.

tests/test_fmi3.py
PeterMeisrimelModelon: Question, (not sure if we should have unit tests for this):
How does it look like for invalid call sequences? 

I recall that a double `terminate()` would simply leave an extra "error log message", but nothing else on the Python side. Which is fine for `terminate()`, but just wondering how it looks like with the rest.

src/pyfmi/fmi3.pyx
```diff
@@ -425,6 +422,11 @@ cdef class FMUModelME3(FMUModelBase3):
         else:
             raise InvalidVersionException('The FMU could not be loaded. This class only supports FMI 3.0 for Model Exchange.')
 
+    def get_identifier(self):
+        if not self._modelId:
```
modelonrobinandersson, line 426: Note that the docstring is automatically fetched from the parent:
```
In [4]: fmu.get_identifier?
Signature: fmu.get_identifier()
Docstring:
Return the model identifier, name of binary model file and prefix in
the C-function names of the model.
Type:      method
```

PeterMeisrimelModelon: Neat, didn't know that!

modelonrobinandersson:
> Looks great thus far. Aside from the comments a few things we should also add here:
> 
> * explicit free_instance function + accompanied appropriate test.
> * fmi3_import_destroy_dllfmu & fmi3_import_free_instance call in `__dealloc__`. See FMI2 for the various occurrences of `_allocated_dll` and `_allocated_fmu`.
> * [Optional] Related to `terminate`, some tests for valid/invalid call sequences, e.g., double termiante is not valid, and the FMU will only "midly" complain, not sure if this should be an error on the PyFMI side though


@PeterMeisrimelModelon  see d9ae5a4 and 6e58cb5

PeterMeisrimelModelon (approved):

PR #298, opened 2025-03-05
Splitting up assimulo_interface.pyx into different FMI versions

PeterMeisrimelModelon:
Since refactoring diffs are awful:
1. `assimulo_interface.pyx`: Remove the unused functions
2. Move the exception to `exceptions.py`
3. split out FMI1/FMI2 into separate files.
4. Cleanup of imports
5. Move the "decide which `FMIODE` class to use" from `fmi_algorithm_drivers.py` to `assimulo_interface.pyx` + simplify it a bit.
6. No "primary" imports were harmed. I.e., importing `FMIODE` from `pyfmi.simulation.assimulo_interface` still works.


src/pyfmi/exceptions.py
```diff
@@ -23,6 +23,12 @@ class FMUException(Exception):
     """
     pass
 
+class FMIModel_Exception(Exception):
```
modelonrobinandersson, line 26: Can we create FMIModelException and extend from this one? Not sure if we can raise a warning for the one with `_` to flag for future removal.

src/pyfmi/simulation/assimulo_interface_fmi2.pyx
```diff
@@ -0,0 +1,859 @@
...
+        self._f_nbr = f_nbr
+        self._g_nbr = g_nbr
+        self._A = None
+
+        self.state_events_use = False
+        if g_nbr > 0:
+            self.state_events_use = True
+        self.time_events_use = True
+
+        #If there is no state in the model, add a dummy state der(y)=0
```
modelonrobinandersson, line 92:
```suggestion
        # If there is no state in the model, add a dummy state der(y)=0
```
and a couple of other places

src/pyfmi/simulation/assimulo_interface_fmi2.pyx
```diff
@@ -0,0 +1,859 @@
...
+    def time_events(self, double t, np.ndarray[double, ndim=1, mode="c"] y, sw=None):
+        """
+        Time event function.
+        """
+        eInfo = self._model.get_event_info()
+
+        if eInfo.nextEventTimeDefined:
+            return eInfo.nextEventTime
+        else:
+            return None
```
modelonrobinandersson, line 363:
```suggestion
        if eInfo.nextEventTimeDefined:
            return eInfo.nextEventTime
        return None
```

modelonrobinandersson (approved):
There is some refactoring/restructuring we could do but let's take that in a separate effort. I think this is good for now.

src/pyfmi/exceptions.py
PeterMeisrimelModelon: Added, also raising the new one; will not affect anyone doing except `FMUModel_Exception`.

I propose we get rid of the old one with PyFMI v3.

modelonrobinandersson: Great!

modelonrobinandersson (approved):

PR #301, opened 2025-04-14
Added many more methods to the FMI3 interface

modelonrobinandersson:


src/pyfmi/fmi3.pxd
```diff
@@ -24,17 +24,38 @@ cimport pyfmi.fmil_import as FMIL
 cimport pyfmi.fmil3_import as FMIL3
 cimport pyfmi.fmi_base as FMI_BASE
 
+cdef class ScalarVariable3:
```
modelonrobinandersson, line 27: Suggestion: Rename to `ModelVariable`, or leave to separate effort.

src/pyfmi/fmi3.pyx
```diff
@@ -568,31 +676,78 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
-    cpdef FMIL3.fmi3_base_type_enu_t get_variable_data_type(self, variable_name) except *:
+    cdef FMIL3.fmi3_base_type_enu_t _get_variable_data_type(self, variable_name) except *:
+        cdef FMIL3.fmi3_import_variable_t* variable
+        variable_name = pyfmi_util.encode(variable_name)
+        cdef char* variablename = variable_name
+
+        variable = FMIL3.fmi3_import_get_variable_by_name(self._fmu, variablename)
+        if variable == NULL:
+            raise FMUException(f"The variable {pyfmi_util.decode(variablename)} could not be found.")
+        return FMIL3.fmi3_import_get_variable_base_type(variable)
```
modelonrobinandersson, line 677: Made internal `_get...` and wrapped below, because Cython cannot convert the type `FMIL3.fmi3_base_type_enu_t` directly to `int` otherwise due to the signature of the function.

src/pyfmi/fmi3.pyx
```diff
@@ -568,31 +676,78 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+        variable_data_type = self._get_variable_data_type(variable_name)
+        return FMI3_Type(int(variable_data_type))
+
+    cdef _add_scalar_variable(self, FMIL3.fmi3_import_variable_t* variable):
+        cdef FMIL3.fmi3_string_t description
 
-        variable = FMIL3.fmi3_import_get_variable_by_name(self._fmu, variablename)
         if variable == NULL:
-            raise FMUException("The variable %s could not be found."%pyfmi_util.decode(variablename))
+            raise FMUException("Unknown variable. Please verify the correctness of the XML file and check the log.")
```
modelonrobinandersson, line 697: This diff looks confusing, I didnt change the error message it just moved down.

src/pyfmi/fmi3.pyx
```diff
@@ -886,6 +1140,178 @@ cdef class FMUModelME3(FMUModelBase3):
...
+        """
+        Returns the nominal values of the continuous states.
+
+        Returns::
+            The nominal values.
+        """
+        cdef FMIL3.fmi3_status_t status
+        cdef np.ndarray[FMIL3.fmi3_float64_t, ndim=1, mode='c'] xn = np.zeros(self._nContinuousStates, dtype=np.double)
+
+        if self._initialized_fmu == 0:
```
modelonrobinandersson, line 1132: Note, new exception for FMI3

src/pyfmi/fmi3.pyx
```diff
@@ -886,6 +1111,174 @@ cdef class FMUModelME3(FMUModelBase3):
...
+            no_set_FMU_state_prior_to_current_point,
+            enter_event_mode,
+            terminate_simulation
+        )
+        self._log_handler.capi_end_callback(self._max_log_size_msg_sent, self._current_log_size)
+
+        self._last_accepted_time = self._get_time()
+        return status
+
+    def completed_integrator_step(self, no_set_FMU_state_prior_to_current_point = True):
```
PeterMeisrimelModelon, line 1180: Any particular reason for the default value here?

Plus should probably add a docstring for the new parameter

src/pyfmi/fmi3.pyx
```diff
@@ -886,6 +1111,174 @@ cdef class FMUModelME3(FMUModelBase3):
...
+        self._log_handler.capi_start_callback(self._max_log_size_msg_sent, self._current_log_size)
+        status = FMIL3.fmi3_import_completed_integrator_step(
+            self._fmu,
+            noSetFMUStatePriorToCurrentPoint,
+            enterEventMode,
+            terminateSimulation
+        )
+        self._log_handler.capi_end_callback(self._max_log_size_msg_sent, self._current_log_size)
+
+        self._last_accepted_time = self._get_time()
```
PeterMeisrimelModelon, line 1209: Should this move after status check? Can we really count this as accepted if the complete_step call failed.

Also, I don't see _last_accepted_time actually being used anywhere, unless I missed something? Might as well just get rid of it?

src/pyfmi/fmi3.pyx
```diff
@@ -886,6 +1111,174 @@ cdef class FMUModelME3(FMUModelBase3):
...
+
+        if np.size(ndx) != self._nContinuousStates:
+            raise FMUException(
+                'Failed to set the new continuous states. ' \
+                'The number of values are not consistent with the number of '\
+                f'continuous states, which are {self._nContinuousStates}.')
+
+        status = self._set_continuous_states_fmil(ndx)
+
+        if status >= 3:
```
PeterMeisrimelModelon, line 1273: Use enum instead of 3?

tests/test_fmi3.py
```diff
@@ -390,6 +465,36 @@ def test_set_array_variable(self):
         with pytest.raises(FMUException, match = err_msg):
             fmu.set("x", np.array([1, 2, 3]))
 
+    def test_get_continuous_states(self):
+        """Test retrieve the continuous states."""
+        fmu_path = FMI3_REF_FMU_PATH / "VanDerPol.fmu"
+        fmu = load_fmu(fmu_path)
+        fmu.initialize()
+        assert all(fmu.continuous_states == [2.0, 0.0])
```
PeterMeisrimelModelon, line 473: The `all` shouldn't be necessary here I believe.

tests/test_fmi3.py
```diff
@@ -233,12 +249,63 @@ def test_get_default_experiment_stop_time(self):
...
 
+    def test_get_states_list(self):
+        """Test retrieving states list and check its attributes. """
+        fmu = load_fmu(FMI3_REF_FMU_PATH / "VanDerPol.fmu")
+        states = fmu.get_states_list()
+
+        assert len(states) == 2
+        x0 = states['x0']
+        x1 = states['x1']
+
```
PeterMeisrimelModelon, line 265: Could add another typecheck for x0, x1 here.

src/pyfmi/fmil3_import.pxd
```diff
@@ -55,17 +61,59 @@ cdef extern from 'fmilib.h':
...
+    cdef enum fmi3_variability_enu_t:
+        fmi3_variability_enu_constant    = 0,
+        fmi3_variability_enu_fixed       = 1,
+        fmi3_variability_enu_tunable     = 2,
+        fmi3_variability_enu_discrete    = 3,
+        fmi3_variability_enu_continuous  = 4,
+        fmi3_variability_enu_unknown     = 5
+
+    cdef enum fmi3_initial_enu_t:
+        fmi3_initial_enu_exact,
```
PeterMeisrimelModelon, line 83: Should we also explicitly state the values here?

tests/test_fmi3.py
```diff
@@ -390,6 +465,36 @@ def test_set_array_variable(self):
...
+            ("String_parameter", FMI3_Type.STRING),
+            ("Binary_input", FMI3_Type.BINARY),
+            ("Enumeration_input", FMI3_Type.ENUM),
+        ]
+    )
+    def test_get_variable_data_type(self, variable_name, expected_datatype):
+        """Test getting variable data types."""
+        fmu_path = FMI3_REF_FMU_PATH / "Feedthrough.fmu"
+        fmu = FMUModelME3(fmu_path, _connect_dll = False)
+        assert fmu.get_variable_data_type(variable_name) is expected_datatype
```
PeterMeisrimelModelon, line 498: Maybe I missed it, but some suggestions for 2 extra tests:

```
states = ...
set(states)
assert states == get(states)
```

```
ders_1 = get_derivatives(...)
states = ...
set(states)
ders_2 = get_derivatives(...)
assert ders_1 != ders_2 
```

Or similar.

PeterMeisrimelModelon (commented):
Nice work! 

Given the size of the MR, a whole lot of minor comments/suggestions.

src/pyfmi/fmi3.pxd
modelonrobinandersson: Good idea, fixed in latest commit

src/pyfmi/fmi3.pxd
```diff
@@ -24,17 +24,38 @@ cimport pyfmi.fmil_import as FMIL
...
+
+
+cdef class EventInfo:
+    cdef public FMIL3.fmi3_boolean_t new_discrete_states_needed
+    cdef public FMIL3.fmi3_boolean_t terminate_simulation
+    cdef public FMIL3.fmi3_boolean_t nominals_of_continuous_states_changed
+    cdef public FMIL3.fmi3_boolean_t values_of_continuous_states_changed
+    cdef public FMIL3.fmi3_boolean_t next_event_time_defined
+    cdef public FMIL3.fmi3_float64_t next_event_time
+
```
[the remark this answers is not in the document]
modelonrobinandersson: Fixed in latest commit

src/pyfmi/fmi3.pyx
```diff
@@ -631,6 +790,64 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
         """ Returns the default experiment tolerance as defined in the XML description. """
         return FMIL3.fmi3_import_get_default_experiment_tolerance(self._fmu)
 
+
+    def get_tolerances(self):
+        """ Returns the relative and absolute tolerances. If the relative tolerance
+            is defined in the XML-file it is used, otherwise a default of 1.e-4 is
```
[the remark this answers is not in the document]
modelonrobinandersson: Yes I also wanted to ask, isnt the docstring wrong, i.e. "otherwise a default of 1.e-4 is used.". As far as I can see I cannot see any default value being used?

src/pyfmi/fmi3.pyx
```diff
@@ -631,6 +790,64 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+
+                rtol --
+                    The relative tolerance.
+        """
+        return self.get_default_experiment_tolerance()
+
+    def get_absolute_tolerances(self):
+        """ Returns the absolute tolerances. They are calculated and returned according to
+            the FMI specification, atol = 0.01*rtol*(nominal values of the
+            continuous states)
```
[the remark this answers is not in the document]
modelonrobinandersson: I agree, but I kept it just because we have it like that in FMI2.

src/pyfmi/fmi3.pyx
```diff
@@ -631,6 +790,64 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+
+            This method should not be called before initialization, since it depends on state nominals.
+
+            Returns::
+
+                atol --
+                    The absolute tolerances.
+        """
+        if self._initialized_fmu == 0:
```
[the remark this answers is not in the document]
modelonrobinandersson: You mean that we should allow `get_absolute_tolerances` before initialization but rely on some other means of computation if the user calls for that?
I think this could also be worth discussing with others more thoroughly.

src/pyfmi/fmi3.pyx
modelonrobinandersson: For this I am not sure, since it is an illegal call I still believe an error is appropriate. Also happy to discuss this together with the absolute tolerance issue above.

src/pyfmi/fmi3.pyx
modelonrobinandersson: I couldn't see why either, but it is the same in FMI2 and I didn't want to change the interface.

tests/test_fmi3.py
modelonrobinandersson: Good catch! Fixed in latest commit.

tests/test_fmi3.py
modelonrobinandersson: Good idea, see the new tests:
def test_get_and_set_states(self)
def test_get_derivatives_with_states_set(self)

src/pyfmi/fmi3.pxd
```diff
@@ -50,12 +71,23 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
     cpdef np.ndarray get_float32(self, valueref)
 
-    cpdef FMIL3.fmi3_value_reference_t get_variable_valueref(self, variablename) except *
-    cpdef FMIL3.fmi3_base_type_enu_t get_variable_data_type(self, variable_name) except *
+    cpdef _get_time(self)
+    cpdef _set_time(self, FMIL3.fmi3_float64_t t)
+
+    cpdef FMIL3.fmi3_value_reference_t get_variable_valueref(self, variable_name) except *
+    cdef FMIL3.fmi3_base_type_enu_t _get_variable_data_type(self, variable_name) except *
+    cdef _add_scalar_variable(self, FMIL3.fmi3_import_variable_t* variable)
```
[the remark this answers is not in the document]
modelonrobinandersson: Good point, I forgot to update this actually. Fixed.

src/pyfmi/fmi3.pyx
modelonrobinandersson: I have moved this now, I agree it didnt seem reasonable as it was in FMI2.

src/pyfmi/fmi3.pyx
modelonrobinandersson: I changed this for now to `FMIL3.fmi3_status_t`, I think in general we should have a separate effort to return enums for all of the functions related to the `FMI3_*`, I would prefer not doing it in this ticket though (or maybe in the end) because of the reason I explained in the other comment about Cython unable to convert Enum instances between the Cython layer and Python.

src/pyfmi/fmil3_import.pxd
modelonrobinandersson: Fixed. I am just worried though assigning `unknown` to a value and it also says "have to be last". For now I think it is fine, but for general enumerations I would be more careful. Because doesnt it mean that the value could change if more members are introduced in-between? Or am I wrong here?
Anyway, added values, see latest commit.

src/pyfmi/fmi3.pyx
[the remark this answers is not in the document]
PeterMeisrimelModelon: Took some digging but here it is: 
https://github.com/modelon-community/fmi-library/blob/master/src/XML/src/FMI3/fmi3_xml_model_description.c#L70

src/pyfmi/fmi3.pxd
PeterMeisrimelModelon: Forgot to push a commit?

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: Ah, I see. 

I suppose we can take this separately later. Wondering if it was possible to move the Python enum to the Cython layer.

src/pyfmi/fmi3.pxd
modelonrobinandersson: Not sure what happened, but now I have pushed the change, see 9b5ed23

src/pyfmi/fmi3.pyx
modelonrobinandersson: I tried that but it seemed to cause complications and it would simply be easier to have it in pure Python. But if you find any nice way to improve it I'm happy to change it.

PR #303, opened 2025-04-23
Adding FMIODE3 class; basic simulation; no events; no result handling

PeterMeisrimelModelon:
General note; I opted for "do-nothing functions" in a lot of places, so we can have a major part of the algorithmic code in place and do not change it later. Hoping this would lead to less TODOs spread in various places.

### DIFF
--- a/src/pyfmi/fmi3.pxd
+++ b/src/pyfmi/fmi3.pxd
@@ -79,6 +79,7 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
     cdef FMIL3.fmi3_base_type_enu_t _get_variable_data_type(self, variable_name) except *
     cpdef get_variable_description(self, variable_name)
     cdef _add_variable(self, FMIL3.fmi3_import_variable_t* variable)
+    cpdef get_variable_unbounded(self, variablename)
 
 
 cdef class FMUModelME3(FMUModelBase3):
@@ -92,6 +93,7 @@ cdef class FMUModelME3(FMUModelBase3):
         FMIL3.fmi3_boolean_t* terminate_simulation
     )
     cdef FMIL3.fmi3_status_t _get_nominal_continuous_states_fmil(self, FMIL3.fmi3_float64_t* xnominal, size_t nx)
+    cdef public object _preinit_nominal_continuous_states
 
 cdef void _cleanup_on_load_error(
     FMIL3.fmi3_import_t* fmu_3,
--- a/src/pyfmi/fmi3.pxd
+++ b/src/pyfmi/fmi3.pxd
@@ -93,7 +93,6 @@ cdef class FMUModelME3(FMUModelBase3):
         FMIL3.fmi3_boolean_t* terminate_simulation
     )
     cdef FMIL3.fmi3_status_t _get_nominal_continuous_states_fmil(self, FMIL3.fmi3_float64_t* xnominal, size_t nx)
-    cdef public object _preinit_nominal_continuous_states
 
 cdef void _cleanup_on_load_error(
     FMIL3.fmi3_import_t* fmu_3,
--- a/src/pyfmi/fmi3.pyx
+++ b/src/pyfmi/fmi3.pyx
@@ -880,6 +880,17 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
         rtol = self.get_relative_tolerance()
         return 0.01*rtol*self.nominal_continuous_states
 
+    cpdef get_variable_unbounded(self, variable_name):
+        """TODO:"""
+        return False
+
+    def get_generation_tool(self):
+        """
+        Return the model generation tool.
+        """
+        cdef FMIL3.fmi3_string_t gen = <FMIL3.fmi3_string_t>FMIL3.fmi3_import_get_generation_tool(self._fmu)
+        return pyfmi_util.decode(gen) if gen != NULL else ""
+
 cdef class FMUModelME3(FMUModelBase3):
     """
     FMI3 ModelExchange model loaded from a dll
@@ -933,6 +944,9 @@ cdef class FMUModelME3(FMUModelBase3):
         self._log_handler.capi_end_callback(self._max_log_size_msg_sent, self._current_log_size)
         if status != FMIL3.fmi3_status_ok:
             raise InvalidFMUException("The FMU could not be instantiated, error retrieving number of continuous states.")
+
+        # State nominals retrieved before initialization
+        self._preinit_nominal_continuous_states = None
 
         if _connect_dll:
             self.instantiate()
@@ -1051,6 +1065,9 @@ cdef class FMUModelME3(FMUModelBase3):
 
         self._initialized_fmu = 1
 
+        # TODO: Ugly workaround
+        self._preinit_nominal_continuous_states = np.array([1.]*self.get_ode_sizes()[0])
+
     def enter_initialization_mode(
         self,
         tolerance_defined=True,
@@ -1144,6 +1161,29 @@ cdef class FMUModelME3(FMUModelBase3):
             raise FMUException("Failed to enter event mode")
 
 
+    def event_update(self, intermediateResult=False):
+        """
+        Updates the event information at the current time-point. If
+        intermediateResult is set to True the update_event will stop at each
+        event iteration which would require to loop until
+        event_info.newDiscreteStatesNeeded == fmiFalse.
+
+        Parameters::
+
+            intermediateResult --
+                If set to True, the update_event will stop at each event
+                iteration.
+                Default: False.
+
+        Example::
+
+            model.event_update()
+
+        Calls the low-level FMI function: TODO
+        """
+        pass
+
+
     cdef FMIL3.fmi3_status_t _get_nominal_continuous_states_fmil(self, FMIL3.fmi3_float64_t* xnominal, size_t nx):
         cdef FMIL3.fmi3_status_t status
         self._log_handler.capi_start_callback(self._max_log_size_msg_sent, self._current_log_size)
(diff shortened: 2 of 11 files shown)


src/pyfmi/fmi3.pyx
```diff
@@ -880,6 +880,17 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
         rtol = self.get_relative_tolerance()
         return 0.01*rtol*self.nominal_continuous_states
 
+    cpdef get_variable_unbounded(self, variable_name):
+        """TODO:"""
+        return False
+
+    def get_generation_tool(self):
```
PeterMeisrimelModelon, line 887: Thought with that one I could get a proper exception for testing sensitivities, but would needs a few more functions. 
Anyways, might as well keep it.

src/pyfmi/fmi3.pyx
```diff
@@ -1228,6 +1268,103 @@ cdef class FMUModelME3(FMUModelBase3):
 
         return values
 
+    def simulate(self,
```
PeterMeisrimelModelon, line 1271: FMI2 copy paste

src/pyfmi/simulation/assimulo_interface_fmi3.pyx
```diff
@@ -0,0 +1,611 @@
...
+# distutils: define_macros=NPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION
+
+"""
+This file contains code for mapping FMI3 FMUs to the Problem specifications
+required by Assimulo.
+"""
+
+import numpy as np
+cimport numpy as np
+
```
PeterMeisrimelModelon, line 27: FMI2 copy paste; commenting out event, result handling and the likes for now.

Plus removed everything dynamic_diagnostics to log.

tests/test_fmi3.py
```diff
@@ -548,6 +548,24 @@ def test_get_variable_data_type(self, variable_name, expected_datatype):
         fmu = FMUModelME3(fmu_path, _connect_dll = False)
         assert fmu.get_variable_data_type(variable_name) is expected_datatype
 
+    def test_simulate(self):
+        """Test basic simulation of an FMU, no result handling."""
```
PeterMeisrimelModelon, line 552: Also check the simulation statistics; they are identical, but I wouldn't want to regression test them here.

src/pyfmi/simulation/assimulo_interface_fmi3.pyx
```diff
@@ -0,0 +1,611 @@
...
+        # If there is no state in the model, add a dummy state der(y)=0
+        if f_nbr == 0:
+            self.y0 = np.array([0.0])
+
+        # Determine the result file name
+        if result_file_name == '':
+            self.result_file_name = model.get_name()+'_result.txt'
+        else:
+            self.result_file_name = result_file_name
```
modelonrobinandersson, line 103: Is it possible to move this out of FMI2 and 3 and into `cExplicit_Problem`? Maybe we can just do something such as `self.set_result_file_name()`
and 
```
def set_result_file_name(self, result_file_name):
  if result_file_name == '':
    self.result_file_name = model.get_name() + '_result.txt'
  else:
    self.result_file_name = result_file_name
```

src/pyfmi/simulation/assimulo_interface_fmi3.pyx
```diff
@@ -0,0 +1,611 @@
...
+        #     self._extra_f_nbr = 0
+        
+        if synchronize_simulation:
+            try:
+                if synchronize_simulation is True:
+                    self._synchronize_factor = 1.0
+                elif synchronize_simulation > 0:
+                    self._synchronize_factor = synchronize_simulation
+                else:
+                    raise InvalidOptionException(f"Setting {synchronize_simulation} as 'synchronize_simulation' is not allowed. Must be True/False or greater than 0.")
```
modelonrobinandersson, line 151: Let's move this message up between `if` and `try` and then we can simply `raise ...(msg)` at both line 151 and 153

src/pyfmi/simulation/assimulo_interface_fmi3.pyx
```diff
@@ -0,0 +1,611 @@
...
+                return 0
+
+            return 1
+        else:
+            return t != self._model.time or (not self._f_nbr == 0 and not (self._model.continuous_states == y).all())
+
+    def rhs(self, double t, np.ndarray[double, ndim=1, mode="c"] y, sw=None):
+        """
+        The rhs (right-hand-side) for an ODE problem.
+        """
```
modelonrobinandersson, line 243: Lets put all of these short docstrings on one line, i.e. `""" foo. """`

src/pyfmi/simulation/assimulo_interface_fmi3.pyx
```diff
@@ -0,0 +1,611 @@
...
+        #     ret_flag =  1 # Tell to reinitiate the solver.
+        
+        # if self._synchronize_factor > 0:
+        #     under_run = solver.t/self._synchronize_factor - (timer()-self._start_time)
+        #     if under_run > 0:
+        #         time.sleep(under_run)
+        
+        # return ret_flag
+
+    def _get_debug_file_object(self):
```
modelonrobinandersson, line 569: Also this should likely be renamed to `get_and_open*` or similar, since the name of the function doesnt state that we actually `open` a resource here (which I would argue is vastly different).

modelonrobinandersson (commented):
This looks great!
I only had a couple of questions and nitpicks on the copy paste from FMI2. Feel free to do as you wish but I thought maybe we can take a minute or two to also make some minor changes while we are at it.

src/pyfmi/simulation/assimulo_interface_fmi3.pxd
```diff
@@ -0,0 +1,55 @@
...
+cimport numpy as np
+import logging
+cimport pyfmi.fmi3 as FMI3
+
+try:
+    import assimulo
+    assimulo_present = True
+except Exception:
+    logging.warning(
+        'Could not load Assimulo module. Check pyfmi.check_packages()')
```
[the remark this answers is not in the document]
PeterMeisrimelModelon: I actually dug a little deeper here; this largely seems to be some left-over bits and pieces.

I suppose the original idea was to have some flexibility for using PyFMI without Assimulo, in particular not being obstructed by failing imports in various places. 
It looks like this track was already abandoned in favor of performance here: https://github.com/modelon-community/PyFMI/commit/4cb0bab4d469e94a8a6ec3222311ef86642e7ab3 (just check the changelog)

So, I suppose we might as well drop the extra `try: ... except: ...`, because it does fail without Assimulo; both building and taking installed PyFMI + Assimulo, removing Assimulo and trying to run the corresponding imports.

src/pyfmi/simulation/assimulo_interface_fmi3.pxd
```diff
@@ -0,0 +1,55 @@
...
+cdef class FMIODE3(cExplicit_Problem):
+    """
+    An Assimulo Explicit Model extended to FMI3 interface.
+    """
+    cdef public int _f_nbr, _g_nbr, _input_activated, _extra_f_nbr, jac_nnz, input_len_names
+    cdef public object _model, problem_name, result_file_name, __input, _A, debug_file_name, debug_file_object
+    cdef public object export, _sparse_representation, _with_jacobian, _logging, _write_header, _start_time
+    cdef public dict timings
+    cdef public np.ndarray y0, input_real_mask, input_other_mask
+    cdef public list input_names, input_real_value_refs, input_other, _logg_step_event
```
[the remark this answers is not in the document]
PeterMeisrimelModelon: I'd leave this to when we start adding support for inputs, but will add a TODO, such that we remember.

src/pyfmi/simulation/assimulo_interface_fmi3.pyx
```diff
@@ -0,0 +1,611 @@
...
+        self._model = model
+        self._adapt_input(input)
+        self.input_names = []
+        self.timings = {"handle_result": 0.0}
+
+        if type(model) == FMI3.FMUModelME3: # if isinstance(model, FMUModelME3):
+            self.model_me3 = model
+            self.model_me3_instance = 1
+        else:
+            self.model_me3_instance = 0
```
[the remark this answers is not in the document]
PeterMeisrimelModelon: Just leave as is for now; Would have to check with Christian on the details again, but in general the rationale for this was to have an optimized pathway such that when using the default implementation, one can completely stick within the compiled Python layer. Whereas if one writes a Python class on top of FMIODE3, this is no longer valid.

src/pyfmi/simulation/assimulo_interface_fmi3.pyx
PeterMeisrimelModelon: `cExplicit_Problem` is in Assimulo; so no on that. 

One could add another intermediate class in PyFMI, but I'd actually prefer to not introduce extra dependencies between the different FMI versions.

src/pyfmi/simulation/assimulo_interface_fmi3.pyx
PeterMeisrimelModelon: Not sure I understand; can you add the suggestion?

The `else` covers e.g., negative values, the `except:` covers nonsense values where e.g., the `synchronize_simulation > 0` evaluation would cause an exception.

Sidenote: I propose we fix this option to be just a float for FMI v3, the unclear type of this is a mess.

src/pyfmi/simulation/assimulo_interface_fmi3.pyx
PeterMeisrimelModelon: Fixed.

src/pyfmi/simulation/assimulo_interface_fmi3.pyx
PeterMeisrimelModelon: Actually, this function(ality) was purely used for the dynamic_diagnostic logging to XML; since we removed that, I could also cleanup this part.

Good that you probed this!

src/pyfmi/fmi3.pyx
```diff
@@ -1144,6 +1161,29 @@ cdef class FMUModelME3(FMUModelBase3):
             raise FMUException("Failed to enter event mode")
 
 
+    def event_update(self, intermediateResult=False):
```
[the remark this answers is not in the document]
modelonrobinandersson: Totally understand, feel free to resolve.

src/pyfmi/simulation/assimulo_interface_fmi3.pyx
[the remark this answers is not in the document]
modelonrobinandersson: Sounds good!

src/pyfmi/fmi3.pxd
```diff
@@ -92,6 +93,7 @@ cdef class FMUModelME3(FMUModelBase3):
         FMIL3.fmi3_boolean_t* terminate_simulation
     )
     cdef FMIL3.fmi3_status_t _get_nominal_continuous_states_fmil(self, FMIL3.fmi3_float64_t* xnominal, size_t nx)
+    cdef public object _preinit_nominal_continuous_states
```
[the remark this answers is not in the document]
PeterMeisrimelModelon: Fixed in https://github.com/modelon-community/PyFMI/pull/303/commits/2f0c4a4823a128f55f7c346307c89a8e0302e5d1

src/pyfmi/simulation/assimulo_interface_fmi3.pxd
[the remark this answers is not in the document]
modelonrobinandersson: Sounds good, resolve

src/pyfmi/simulation/assimulo_interface_fmi3.pyx
modelonrobinandersson: Oh, I missed that (just resolve)

src/pyfmi/simulation/assimulo_interface_fmi3.pyx
modelonrobinandersson: I could be more clear, I just mean that we can move the message to a variable to reduce the duplicate exact string, i.e.
```
msg = "..."
if ...
 raise Exception1(msg)
except ...
 raise Exception2(msg)
```
(since we raise the exact same error message)

src/pyfmi/fmi3.pyx
```diff
@@ -287,7 +287,7 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
         # Connect the DLL
         if _connect_dll:
             self._log_handler.capi_start_callback(self._max_log_size_msg_sent, self._current_log_size)
-            status = FMIL3.fmi3_import_create_dllfmu(self._fmu, self._fmu_kind, NULL, FMIL3.fmi3_log_forwarding)
+            status = FMIL3.fmi3_import_create_dllfmu(self._fmu, self._fmu_kind, NULL, NULL)
```
modelonrobinandersson, line 290: oh is this what caused it to log to stdout?

PR #304, opened 2025-04-25
Possible fix for erroneous @Diagnostics.cpu_time returns from get_var…

PeterMeisrimelModelon:
…iables_data

### DIFF
--- a/src/common/io.py
+++ b/src/common/io.py
@@ -1546,7 +1546,7 @@ class ResultDymolaBinary(ResultDymola):
                 name == f'{DIAGNOSTICS_PREFIX}cpu_time'
             ):
             return Trajectory(
-                time, np.cumsum(self.get_variable_data(f'{DIAGNOSTICS_PREFIX}cpu_time_per_step').x[start_index:stop_index]))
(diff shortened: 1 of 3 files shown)


src/common/io.py
```diff
@@ -1546,7 +1546,7 @@ def _get_variable_data_as_trajectory(self,
                 name == f'{DIAGNOSTICS_PREFIX}cpu_time'
             ):
             return Trajectory(
-                time, np.cumsum(self.get_variable_data(f'{DIAGNOSTICS_PREFIX}cpu_time_per_step').x[start_index:stop_index]))
+                time, np.cumsum(self.get_variable_data(f'{DIAGNOSTICS_PREFIX}cpu_time_per_step').x)[start_index:stop_index])
```
PeterMeisrimelModelon, line 1549: This means we always get the full trajectory for this particular variable. 
Might not be the most efficient, but is robust; I fear anything else might get fairly complicated.

Also, this is the only cumulative variable that requires this extra treatment

modelonrobinandersson: Oh good catch!

modelonrobinandersson (approved):

PR #305, opened 2025-04-28
Allow to choose fmilib library name

jschueller:
This allows to override the default.
Also change the default to link to shared libs on linux too for consistency with Windows, also as distros have a strong preference for them.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -99,6 +99,7 @@ else:
 
 static = False
 debug_flag = False
+fmil_name = "fmilib_shared"
 fmilib_shared = ""
 copy_gcc_lib = False
 gcc_lib = None
@@ -120,6 +121,9 @@ for x in sys.argv[1:]:
         incdirs = [os.path.join(x[12:],'include')]
         libdirs = [os.path.join(x[12:],'lib'), os.path.join(x[12:],'lib64')]
         bindirs = [os.path.join(x[12:],'bin')]
+        copy_args.remove(x)
+    if not x.find('--fmil-name'):
+        fmil_name = x[12:]
         copy_args.remove(x)
     if not x.find('--copy-libgcc'):
         if x[14:].upper() == "TRUE":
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -99,6 +99,7 @@
 
 static = False
 debug_flag = False
+fmil_name = "fmilib_shared" if sys.platform.startswith("win") else "fmilib" #If windows shared, else static
```
modelonrobinandersson, line 102: @jschueller I saw your comment here https://github.com/modelon-community/fmi-library/issues/155#issuecomment-2833547365 as well. I wonder if we should just change the build system to always use static (or shared) for both platforms. I think the reason why this exists currently in the code is an old legacy reason where we had problems creating a static lib on Windows.

jschueller: yes, lets use fmilib_shared by default everywhere
I think we should keep this option for retrocompatibility in case we change fmillib default lib name
(I have in mind that we should avoid creating both shared and static lib in fmilib but allow either one with the same library name, ie this might change the default name to just "fmilib")

jschueller:
I cannot test windows stuff easily, unless a CI test is added.

jschueller:
I think its copying the windows dll next to the python module as there is not many ways to set dlls search paths,
so it might be required anyways

modelonrobinandersson:
> I think its copying the windows dll next to the python module as there is not many ways to set dlls search paths, so it might be required anyways

Ill see if I have to investigate this later this week or the next.

jschueller:
@modelonrobinandersson I think this can be merged as is

modelonrobinandersson:
> @modelonrobinandersson I think this can be merged as is

What was the conclusion again regarding using static/shared on linux? I see the last change in the PR will change the default behavior on Linux: since we are removing this conditional here
```
ext_list[i].libraries = ["fmilib_shared"] if sys.platform.startswith("win") else ["fmilib"] #If windows shared, else static
```

jschueller:
we will use fmilib_shared by default for all platforms instead of windows-only

jschueller:
@modelonrobinandersson is there a new version planned ? it would be cool to have this change and the cython fix in a new patch release

jschueller:
I dont have write access

modelonrobinandersson:
> I dont have write access

@jschueller oh sorry I wasn't aware. Can you let me know which exact commits you want me to cherry-pick over and I could help get it in place

jschueller:
relevant commits are 44b40aa and 028f9d8

modelonrobinandersson:
@jschueller https://github.com/modelon-community/PyFMI/releases/tag/PyFMI-2.17.1

PR #308, opened 2025-05-01
Implemented get_model_variables

modelonrobinandersson:


src/pyfmi/fmi3.pyx
```diff
@@ -675,6 +678,155 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+            filter --
+                Filter the variables using a unix filename pattern
+                matching (filter="*der*"). Can also be a list of filters
+                See http://docs.python.org/2/library/fnmatch.html.
+                Default: None
+
+        Returns::
+
+            Dict with variable name as key and a pyfmi.fmi3.FMI3ModelVariable class as value.
+        """
```
modelonrobinandersson, line 731: When reviewing, might be worth to also look at the FMI2 implementation, I did a lot of clean-up as well and removed variables that looked redundant to me.

src/pyfmi/fmi3.pyx
```diff
@@ -675,6 +678,155 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+    #@enable_caching TODO: Doesnt work using enums with members with C-types, lets discuss
+    def get_model_variables(self,
+        type: Optional[FMI3_Type] = None,
+        include_alias: bool = True,
+        causality: Optional[FMI3_Causality] = None,
+        variability: Optional[FMI3_Variability] = None,
+        only_start: bool = False,
+        only_fixed: bool = False,
+        filter = None,
+        _as_list: bool = False) -> Dict[str, FMI3ModelVariable]:
```
modelonrobinandersson, line 691: Can we remove `_as_list`, is only used in one place in `io.py` for FMI2, and can easily be adjusted.

src/pyfmi/fmi3.pyx
```diff
@@ -675,6 +678,155 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+                )
+
+        FMIL3.fmi3_import_free_variable_list(variable_list)
+
+        if _as_list:
+            return list(variable_dict.values())
+        else:
+            return variable_dict
+
+    def get_input_list(self):
```
modelonrobinandersson, line 817: Added this for testing (since it relies on get_model_variables)

src/pyfmi/fmi3.pyx
```diff
@@ -689,6 +841,14 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
         variable_data_type = self._get_variable_data_type(variable_name)
         return FMI3_Type(int(variable_data_type))
 
+    cdef _get_variable_description(self, FMIL3.fmi3_import_variable_t* variable):
```
modelonrobinandersson, line 844: To reduce code duplication

src/pyfmi/fmi3.pyx
```diff
@@ -675,6 +678,155 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
             raise FMUException(f"The variable {pyfmi_util.decode(variablename)} could not be found.")
         return FMIL3.fmi3_import_get_variable_base_type(variable)
 
+
+    #@enable_caching TODO: Doesnt work using enums with members with C-types, lets discuss
```
PeterMeisrimelModelon, line 682: Was curious, can work around this by using IntEnum and wrapping it as follows:

```
    def get_model_variables(self,
        type: Optional[FMI3_Type] = None,
        include_alias: bool = True,
        causality: Optional[FMI3_Causality] = None,
        variability: Optional[FMI3_Variability] = None,
        only_start: bool = False,
        only_fixed: bool = False,
        filter = None,
        _as_list: bool = False) -> Dict[str, FMI3ModelVariable]:

        return self._get_model_variables(
            type = int(type) if type is not None else None,
            include_alias = include_alias,
            causality = int(causality) if causality is not None else None,
            variability = int(variability) if variability is not None else None,
            only_start = only_start,
            only_fixed = only_fixed,
            filter = filter,
            _as_list = _as_list
        )

    @enable_caching
    def _get_model_variables(self,
        type: Optional[FMI3_Type] = None,
        include_alias: bool = True,
        causality: Optional[FMI3_Causality] = None,
        variability: Optional[FMI3_Variability] = None,
        only_start: bool = False,
        only_fixed: bool = False,
        filter = None,
        _as_list: bool = False) -> Dict[str, FMI3ModelVariable]:
```

A few tests fail, but the function calls works.

I think we should maybe revisit the enum stuff; While I really like it, I still think it should be somewhat optional.

src/pyfmi/fmi3.pyx
```diff
@@ -675,6 +678,155 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+        cdef FMIL3.fmi3_import_variable_list_t*   variable_list
+        cdef FMIL.size_t                          variable_list_size
+        cdef FMIL3.fmi3_value_reference_t         value_ref
+        cdef FMIL3.fmi3_base_type_enu_t           data_type
+        cdef FMIL3.fmi3_variability_enu_t         data_variability
+        cdef FMIL3.fmi3_causality_enu_t           data_causality
+        cdef FMIL3.fmi3_initial_enu_t             data_initial
+
+        cdef int has_start = 0
+        # TODO: Can we rename the keyword arg 'filter' to variable_filter?
```
PeterMeisrimelModelon, line 742: As much as I dislike the use of keywords as as parameter names here, we may want to keep it for consistency and revisit for a 3.x release, when we can break things.

src/pyfmi/fmi3.pyx
```diff
@@ -675,6 +678,155 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+            name             = pyfmi_util.decode(FMIL3.fmi3_import_get_variable_name(variable))
+            value_ref        = FMIL3.fmi3_import_get_variable_vr(variable)
+            description      = self._get_variable_description(variable)
+            data_causality   = FMIL3.fmi3_import_get_variable_causality(variable)
+            data_initial     = FMIL3.fmi3_import_get_variable_initial(variable)
+            data_type        = FMIL3.fmi3_import_get_variable_base_type(variable)
+            #If only variables with start are wanted, check if the variable has start
+
+            # TODO: Discuss if we want to support also regular integers as inputs
+            if isinstance(variable_type, FMI3_Type) and (data_type != variable_type.value):
```
PeterMeisrimelModelon, line 783: Not sure if Cython will be smart enough to only evaluate these once? Might want to explicitly calculate it only once like we did in FMI2.

src/pyfmi/fmi3.pyx
```diff
@@ -675,6 +678,155 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+                continue
+            if isinstance(variability, FMI3_Variability) and (data_variability != variability.value):
+                continue
+            if isinstance(causality, FMI3_Causality) and (data_causality != causality.value):
+                continue
+
+            if variable_filter:
+                filter_list = self._convert_filter(variable_filter)
```
PeterMeisrimelModelon, line 791: I think this absolutely needs to be moved out of the loop to avoid converting the filter list for each variable.

tests/test_fmi3.py
```diff
@@ -358,6 +358,229 @@ def test_get_description_variable_not_found(self):
...
 
+    def test_get_model_variables(self):
+        """ Test get_model_variables with default arguments. """
+        fmu_path = FMI3_REF_FMU_PATH / "VanDerPol.fmu"
+        fmu = load_fmu(fmu_path)
+        variables = fmu.get_model_variables()
+
+        assert len(variables) == 6
+
+        v = variables['time']
```
PeterMeisrimelModelon, line 369: missing asserts here?

src/pyfmi/fmi3.pyx
modelonrobinandersson: Thanks! Updated.

src/pyfmi/fmi3.pyx
modelonrobinandersson: Fixed (will account for the removal in my other PR for result handling)

src/pyfmi/fmi3.pyx
modelonrobinandersson: I agree.
I thought also we could do this:
current:
```
def foo(self, ..., type = None, ...)
```
new:
```
def foo(self, ..., variable_type = None, ..., type = None):
if type is not None:
  # inform user that 'type' is deprecated and they should use variable_type
  # check if user specified both  and raise error (no user will a be affected by this yet, but could in the future)
   variable_type = type
```
This way, the order of the argument has not changed, and if the user has specified `type=...` to something, then they will see the warning. Thoughts? I'm also fine to just leave it for now but wanted to share my idea.

src/pyfmi/fmi3.pyx
modelonrobinandersson: I think this should be fine, it is also `get_model_variables` which is not used repeatedly during simulation and we had`if ... is not None` in the FMI2 specification so I think the difference is minimal if anything.

src/pyfmi/fmi3.pyx
modelonrobinandersson: Good catch, fixed!

tests/test_fmi3.py
modelonrobinandersson: Good catch!

src/pyfmi/fmi3.pyx
```diff
@@ -675,6 +678,168 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+
+        variable_list      = FMIL3.fmi3_import_get_variable_list(self._fmu, 0)
+        variable_list_size = FMIL3.fmi3_import_get_variable_list_size(variable_list)
+
+        if variable_filter:
+            filter_list = self._convert_filter(variable_filter)
+
+        user_specified_type        = isinstance(variable_type, int)
+        user_specified_variability = isinstance(variability, int)
+        user_specified_causality   = isinstance(causality, int)
```
modelonrobinandersson, line 773: @PeterMeisrimelModelon  d8aaeea

src/pyfmi/fmi3.pyx
modelonrobinandersson: See d8aaeea

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: Based on the inputs we know it is `int` or `None`, I wonder if it was faster to instead check  to do `user_specified_type = variable_type is not None`. 

Feel free to test/ignore/fix/resolve.

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: Not sure if there is much point in introducing this as deprecated for FMI3 to begin with. As I see it, we want this to be gone, yet be consistent with FMI1 & 2, so these are the problematic parts.

I think this would be a great solution if we wanted to move away from this as a whole, including FMI1 & 2.

PeterMeisrimelModelon (approved):

PR #309, opened 2025-05-05
Draft: Result handling for FMI 3

modelonrobinandersson:


tests/test_fmi3_sim.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python
```
modelonrobinandersson, line 1: Since `test_fmi3.py` is getting large I decided to add simulation related tests here.

src/pyfmi/fmi_util.pyx
```diff
@@ -205,14 +259,32 @@ cpdef prepare_data_info(np.ndarray[int, ndim=2] data_info, list sorted_vars, lis
...
-                                    np.array(diagnostics_param_values).astype(float)),
-                                    axis = 0
-                                )
-                    )
+    if is_fmi3:
+        data = np.append(
+            model.time,
+            np.concatenate(
+                (model.get_float64(param_float64),
+                 model.get_float64(param_float32),
```
PeterMeisrimelModelon, line 267: This would fail if we had actual float32 variables?

src/pyfmi/fmi_util.pyx
```diff
@@ -944,6 +1020,23 @@ cdef class DumpData:
...
                 self.model_me2._get_boolean(self.bool_var_ref, self.bool_size, self.bool_var_tmp)
                 self.dump_data(self.bool_var_tmp)
+        elif self.model_me3_instance:
+
+            self.time_tmp[0] = self.model_me3.time
+            self.dump_data(self.time_tmp)
+
+            if self.real_size > 0:
+                # TODO: Do we want a get_float64_by_list also for performance reasons?
+                r = self.model_me3.get_float64(self.real_var_ref)
```
PeterMeisrimelModelon, line 1030: Don't we need a separation of real_var_ref into float32, 64?

src/pyfmi/fmi_util.pyx
```diff
@@ -944,6 +1020,23 @@ cdef class DumpData:
...
+                self.dump_data(r)
+
+            if self.int_size > 0:
+                i = self.model_me3.get_int64(self.int_var_ref).astype(float)
```
PeterMeisrimelModelon, line 1034: Same things for ints here?

tests/test_fmi3.py
```diff
@@ -581,6 +580,28 @@ def test_get_input_list(self):
 
         assert expected == list(inputs.keys())
 
+    def test_get_int32_and_64(self):
```
PeterMeisrimelModelon, line 583: Possibly split into 2 different tests; maybe parameterize? Will likely need/want tests for the other int types.

tests/test_fmi3.py
```diff
@@ -581,6 +580,28 @@ def test_get_input_list(self):
...
+        """ Test get int32 and int64. """
+        fmu_path = FMI3_REF_FMU_PATH / "Feedthrough.fmu"
+        fmu = load_fmu(fmu_path)
+        int64_vals = fmu.get_int64([23, 24])
+        int32_vals = fmu.get_int32([19, 20])
+
+        np.testing.assert_equal(np.array([0, 0]), int64_vals)
```
PeterMeisrimelModelon, line 590: A thought; maybe we want to implement the setters as well and properly test:
1. Getting/setting works e.g., for int64s for values that would not fit into an int32
2. Connected to that also the practical limitations of storing results for these?

tests/test_fmi3_sim.py
```diff
@@ -0,0 +1,70 @@
...
+    def test_simulate(self):
+        """Test simulate VDP model and verify the integrity of the results. """
+        fmu = load_fmu(FMI3_REF_FMU_PATH / "VanDerPol.fmu")
+        results = fmu.simulate()
+
+        assert results['x0'][0] == 2.0
+        assert results['x1'][0] == 0.0
+        assert results['x0'][-1] == pytest.approx( 2.00814337)
+        assert results['x1'][-1] == pytest.approx(-0.04277047)
+        np.testing.assert_equal(results['mu'], np.ones(501))
```
PeterMeisrimelModelon, line 43: Could we change this one as to not regression test the default NCP?

tests/test_fmi3_sim.py
```diff
@@ -0,0 +1,70 @@
...
+        """Test simulate VDP model and check accessible data. """
+        fmu = load_fmu(FMI3_REF_FMU_PATH / "VanDerPol.fmu")
+        results = fmu.simulate()
+
+        expected_variables = ['time', 'x0', 'der(x0)', 'x1', 'der(x1)', 'mu']
+
+        assert results.keys() == expected_variables
+
+        for v in expected_variables:
+            assert len(results[v]) == 501
```
PeterMeisrimelModelon, line 55: I don't think we need to check the lengths here.

tests/test_fmi3_sim.py
```diff
@@ -0,0 +1,70 @@
...
+
+        opts['ncp'] = 600
+        results = fmu.simulate(options = opts)
+
+        expected_variables = ['time', 'x0', 'der(x0)', 'x1', 'der(x1)', 'mu']
+
+        assert results.keys() == expected_variables
+
+        for v in expected_variables:
+            assert len(results[v]) == 601
```
PeterMeisrimelModelon, line 70:
```suggestion
        assert all(len(result[v]) == 601 for v in expected_variables)
```

src/pyfmi/fmi_util.pyx
modelonrobinandersson: Typo, has been fixed now

src/pyfmi/fmi_util.pyx
modelonrobinandersson: I think you are right and same for `int32|64`, I will fix that tomorrow.

src/pyfmi/fmi_util.pyx
modelonrobinandersson: Correct, leaving this open.

tests/test_fmi3.py
modelonrobinandersson: Split it into 2 for now, I think we can do the parametrization when we add the other tests later.

tests/test_fmi3.py
modelonrobinandersson: Ill create a separate issue for this.

tests/test_fmi3_sim.py
modelonrobinandersson: Good idea, fixed

tests/test_fmi3_sim.py
modelonrobinandersson: Initially I was thinking the same, but then I realized for example there is a bug right now for **diagnostics data** such that if you enable it, you only get 1 point of data (for both non-diagnostics and diagnostics). Therefore I thought it could be reasonable with some sanity check, but, I went ahead and removed it.

tests/test_fmi3_sim.py
modelonrobinandersson: Nice! Fixed.

src/pyfmi/fmi_util.pyx
modelonrobinandersson: @PeterMeisrimelModelon a lot of changes in commit https://github.com/modelon-community/PyFMI/pull/309/commits/46c479850f24e90480ff246216031f6a44efa905

src/pyfmi/fmi_util.pxd
```diff
@@ -49,6 +50,16 @@ ELSE:
     cdef inline long long os_specific_ftell(FILE *stream):
         return ftello(stream)
 
+cdef class DumpDataFMI3:
+    cdef np.ndarray time_tmp
+    # TODO: Would make more sense if 'model' is declared as FMI3.FMUModelME3
```
PeterMeisrimelModelon, line 55: Why not; this seems to work for FMI2?

src/pyfmi/fmi_util.pyx
```diff
@@ -151,20 +154,19 @@ cpdef prepare_data_info(np.ndarray[int, ndim=2] data_info, list sorted_vars, lis
             data_info[0,i] = last_data_matrix
             data_info[1,i] = alias*last_index
         else:
-            variability = var.variability
             last_vref   = var.value_reference
-            data_type   = var.type
```
PeterMeisrimelModelon, line 156: Not sure if this change improves things? It seems sensible that these computations are broken out?

src/pyfmi/fmi_util.pyx
```diff
@@ -151,20 +154,19 @@ cpdef prepare_data_info(np.ndarray[int, ndim=2] data_info, list sorted_vars, lis
...
                 last_data_matrix = 1
                 index_fixed = index_fixed + 1
                 last_index = index_fixed
 
-                if data_type == _FMI_REAL:
+                # Make make sure we add the parameters to the correct list of value references
+                if var.type == FMI_REAL:
                     param_real.append(last_vref)
-                elif data_type == _FMI_INTEGER or data_type == _FMI_ENUMERATION:
+                elif (var.type == FMI_INTEGER) or (var.type == FMI_ENUMERATION):
```
PeterMeisrimelModelon, line 167: I see you changed from the `_FMI_TYPES` to `FMI_TYPE` here; If this doesn't cause any issues, we should maybe fully apply in this function and remove the definitions up top.

src/pyfmi/fmi_util.pyx
```diff
@@ -205,17 +208,130 @@ cpdef prepare_data_info(np.ndarray[int, ndim=2] data_info, list sorted_vars, lis
...
+        #else:
+        #    alias = 1
+        alias = 1
+
+        if last_vref == var.value_reference:
+            data_info[0,i] = last_data_matrix
+            data_info[1,i] = alias*last_index
+        else:
+            last_vref = var.value_reference
+            is_fixed_or_const = var.variability in (FMI3_Variability.FIXED,  FMI3_Variability.CONSTANT)
```
PeterMeisrimelModelon, line 274:
```suggestion
            is_fixed_or_const = var.variability in (FMI3_Variability.FIXED, FMI3_Variability.CONSTANT)
```

src/pyfmi/fmi_util.pyx
```diff
@@ -205,17 +208,130 @@ cpdef prepare_data_info(np.ndarray[int, ndim=2] data_info, list sorted_vars, lis
...
+    cdef dict variables = {
+        FMI3_Type.FLOAT64: [],
+        FMI3_Type.FLOAT32: [],
+        FMI3_Type.INT64: [],
+        FMI3_Type.INT32: [],
+        FMI3_Type.BOOL: [],
+        # TODO: Additional types
+    }
+    # "To set or inquire variables of type Enumeration, fmi3SetInt64 and fmi3GetInt64 must be used"
+    params[FMI3_Type.ENUM] = params[FMI3_Type.INT64]
```
PeterMeisrimelModelon, line 252: Not sure I get this; aren't these empty for now?
In the current format we'd still store any enum parameters in params[ENUM]?

src/pyfmi/fmil3_import.pxd
```diff
@@ -31,6 +30,8 @@ cdef extern from 'fmilib.h':
     ctypedef bool     fmi3_boolean_t
     ctypedef double   fmi3_float64_t
     ctypedef float    fmi3_float32_t
+    ctypedef int      fmi3_int64_t
```
PeterMeisrimelModelon, line 33: This doesn't quite feel right?

tests/test_fmi3.py
```diff
@@ -581,6 +580,36 @@ def test_get_input_list(self):
 
         assert expected == list(inputs.keys())
 
+    def test_get_int32(self):
+        """ Test get int32. """
+        fmu_path = FMI3_REF_FMU_PATH / "Feedthrough.fmu"
+        fmu = load_fmu(fmu_path)
+        int32_vals = fmu.get_int32([19, 20])
+
+        np.testing.assert_equal(np.array([0, 0], dtype=np.int32), int32_vals)
```
PeterMeisrimelModelon, line 589: Consideration for these tests; should be also implement the setters and test with proper values. E.g., for the int64 test actually test a value that would not fit an int32?

tests/test_fmi3_sim.py
```diff
@@ -0,0 +1,96 @@
...
+    @pytest.mark.parametrize("ref_fmu, nbr_of_vars",
+        [
+            (FMI3_REF_FMU_PATH / "Dahlquist.fmu",    4),
+            (FMI3_REF_FMU_PATH / "Feedthrough.fmu",  19),
+            (FMI3_REF_FMU_PATH / "Resource.fmu",     2),
+            (FMI3_REF_FMU_PATH / "Stair.fmu" ,       2),
+            (FMI3_REF_FMU_PATH / "VanDerPol.fmu",    6),
+        ]
+    )
+    def test_simulate_reference_fmus(self, ref_fmu, nbr_of_vars):
```
PeterMeisrimelModelon, line 89: Not sure I see the point in testing many different FMUs here.

I'd suggest maybe just using the Feedthrough.fmu + explicitly assert a given variable is in the result or not, since that one contains all types.

tests/test_fmi3_sim.py
```diff
@@ -0,0 +1,96 @@
...
+        assert results.keys() == expected_variables
+
+        assert all(len(results[v]) == 601 for v in expected_variables)
+
+    def test_simulate_unsupported_result_handler(self):
+        """Verify unsupported result handlers raises an exception"""
+        fmu = load_fmu(FMI3_REF_FMU_PATH / "VanDerPol.fmu")
+        opts = fmu.simulate_options()
+
+        for rh in ['file', 'memory', 'csv']:
```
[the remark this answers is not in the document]
modelonrobinandersson: Great idea, fixed

tests/test_fmi3_sim.py
modelonrobinandersson: Great point, fixed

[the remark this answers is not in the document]
PeterMeisrimelModelon: Sure, could be a good measure of progress.

I don't think testing the number of variables in the result should be regression tested then though.

Might make sense to test with respect to a reference result at final time from FMI2, don't need to do that now though.

PeterMeisrimelModelon: Minor comment on the fix; Note that this also tests the order right now. Might want to use sets instead? 

Seems like the kind of thing to stumble over when adding more variables.

src/pyfmi/fmi_util.pyx
PeterMeisrimelModelon: Ah, right.

How about explicitly copying the enum paramters & variables into corresponding int64 places after all variables/parameter have been sorted?

src/pyfmi/fmi_util.pyx
PeterMeisrimelModelon: Yes, fully getting rid of the _ prefix; not sure I see the point in them.

src/pyfmi/fmi_util.pxd
PeterMeisrimelModelon: From a brief look on FMI2; `model` is a also a generic `public object` there, but there is the additional explicit `model_me2` there, with assignment behind an explicit type check. 

We may want to use a similar format here; I don't fully understand the practical implications, but it was motivated by performance if I recall correctly. Could also postpone this until later when we evaluate performance for FMI3.
Either way, maybe could add a comment as reminder?

PeterMeisrimelModelon (commented):
A few minor comments left!

src/pyfmi/fmi_util.pyx
modelonrobinandersson: I changed this again, I don't think the old approach was correct after thinking about it for a while. Initially that would cause some variables to be written to the result twice. Now instead with my latest commit:

1. Individual lists for int64 and enum
2. When concatenating the huge array with all the parameters I now simply added `model.get_int64(params[FMI3_Type.ENUM]).astype(float),` in the end.
3. I also temporarily added a `print(...)` in the function (not committed) `DumpDataFMI3.save_point` to verify only enums were written to result at the time and now I think it behaves as expected.

src/pyfmi/fmi_util.pyx
PeterMeisrimelModelon: I agree that it is more readable, but not sure to which degree the compilation of Python code can optimize this. Namely the question that these class attributes do not have a constant variability, yet are not explicitly changed in between.

(Feel free to decide & resolve)

modelonrobinandersson: Hmm I'll leave this as it is now and I think when we do a wider performance investigation it will show up if significant. (I think we should benchmark and compare with detailed output between FMI2 and FMI3 in a later stage.

tests/test_fmi3_sim.py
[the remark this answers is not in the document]
modelonrobinandersson: @PeterMeisrimelModelon this test has been updated now.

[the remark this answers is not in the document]
modelonrobinandersson: Test has been updated

PeterMeisrimelModelon (approved):
Good to merge; a rebase should fix the GitHub Action build issue.

PR #312, opened 2025-05-09
Adding FMI3 alias API

PeterMeisrimelModelon:
I think there is an obvious need for some convenience functionality in FMIL around aliases. Can use this here as a basis for that discussion.

### DIFF
--- a/src/pyfmi/fmi3.pyx
+++ b/src/pyfmi/fmi3.pyx
@@ -1270,7 +1270,7 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
     def get_variable_alias(self, variable_name):
         """
         Return a dict of all alias variables belonging to the provided variable
-        where the key are the names and the value indicati whether the variable
+        where the key are the names and the value indicates if the variable
         is an alias or not.
 
         Parameters::
--- a/src/pyfmi/fmi3.pyx
+++ b/src/pyfmi/fmi3.pyx
@@ -1079,9 +1079,6 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
             raise FMUException("The variable %s could not be found." % variable_name)
 
     cdef _add_variable(self, FMIL3.fmi3_import_variable_t* variable):
-        """Not to be used for creating aliased."""
-        # XXX: Alias would require having the desired name as input
-
         if variable == NULL:
             raise FMUException("Unknown variable. Please verify the correctness of the XML file and check the log.")
 
--- a/src/pyfmi/fmi3.pyx
+++ b/src/pyfmi/fmi3.pyx
@@ -101,7 +101,7 @@ cdef void importlogger3(FMIL.jm_callbacks* c, FMIL.jm_string module, FMIL.jm_log
 
 cdef class FMI3ModelVariable:
     """ Class defining data structure based on the XML elements of ModelVariables. """
-    def __init__(self, name, value_reference, data_type, description, variability, causality, initial):
+    def __init__(self, name, value_reference, data_type, description, variability, causality, alias, initial):
         self._name            = name
         self._value_reference = value_reference
         self._type            = data_type
@@ -109,7 +109,7 @@ cdef class FMI3ModelVariable:
         self._variability     = variability
         self._causality       = causality
         self._initial         = initial
-        self._alias           = 1 # dummy for now
+        self._alias           = alias
 
     def _get_name(self):
         return self._name
@@ -888,14 +888,18 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
         only_start,
         only_fixed,
         filter):
-        cdef FMIL3.fmi3_import_variable_t*        variable
-        cdef FMIL3.fmi3_import_variable_list_t*   variable_list
-        cdef FMIL.size_t                          variable_list_size
-        cdef FMIL3.fmi3_value_reference_t         value_ref
-        cdef FMIL3.fmi3_base_type_enu_t           data_type
-        cdef FMIL3.fmi3_variability_enu_t         data_variability
-        cdef FMIL3.fmi3_causality_enu_t           data_causality
-        cdef FMIL3.fmi3_initial_enu_t             data_initial
+        cdef FMIL3.fmi3_import_variable_t*           variable
+        cdef FMIL3.fmi3_import_variable_list_t*      variable_list
+        cdef FMIL.size_t                             variable_list_size
+        cdef FMIL3.fmi3_value_reference_t            value_ref
+        cdef FMIL3.fmi3_base_type_enu_t              data_type
+        cdef FMIL3.fmi3_variability_enu_t            data_variability
+        cdef FMIL3.fmi3_causality_enu_t              data_causality
+        cdef FMIL3.fmi3_initial_enu_t                data_initial
+
+        cdef FMIL3.fmi3_import_alias_variable_list_t* alias_list
+        cdef FMIL.size_t                              alias_list_size
+        cdef FMIL3.fmi3_import_alias_variable_t*      alias_var
 
         cdef int has_start = 0
         # TODO: Can we rename the keyword arg 'filter' to variable_filter?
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         python-version: '3.11.x'
     - name: Setup Python
       run: |
-        python3 -m pip install Cython numpy scipy matplotlib pytest setuptools==69.1.0
+        python3 -m pip install Cython==3.0.12 numpy scipy matplotlib pytest setuptools==69.1.0
```
PeterMeisrimelModelon, line 19: Testing only

src/pyfmi/fmi3.pyx
```diff
@@ -794,7 +803,7 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
             data_causality   = FMIL3.fmi3_import_get_variable_causality(variable)
             data_initial     = FMIL3.fmi3_import_get_variable_initial(variable)
             data_type        = FMIL3.fmi3_import_get_variable_base_type(variable)
-            #If only variables with start are wanted, check if the variable has start
+            # If only variables with start are wanted, check if the variable has start
 
             # TODO: Discuss if we want to support also regular integers as inputs
```
modelonrobinandersson, line 808: By the way we can remove this comment "# TODO: ...", I realize I forgot to remove it in the PR where I made the Enums into IntEnum.

src/pyfmi/fmi3.pyx
```diff
@@ -101,19 +101,24 @@ cdef void importlogger3(FMIL.jm_callbacks* c, FMIL.jm_string module, FMIL.jm_log
...
         self._initial         = initial
+        self._alias           = alias
 
     def _get_name(self):
         return self._name
     name = property(_get_name)
 
+    def _get_alias(self):
+        return self._alias
+    alias = property(_get_alias)
```
modelonrobinandersson, line 120: Should we rename this to `is_alias` considering its a bool?

src/pyfmi/fmi3.pyx
```diff
@@ -855,7 +883,15 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
         return FMI3_Type(int(variable_data_type))
 
     cdef _get_variable_description(self, FMIL3.fmi3_import_variable_t* variable):
-        cdef FMIL3.fmi3_string_t desc = FMIL3.fmi3_import_get_variable_description(variable)
+        cdef FMIL3.fmi3_string_t desc = <FMIL3.fmi3_string_t>FMIL3.fmi3_import_get_variable_description(variable)
```
modelonrobinandersson, line 886: Did this really work? This is what crashed for me a few months ago which resulted in the previous implementation. I mean adding `my_type` to `foo = <my_type>bar` crashed.

src/pyfmi/fmi3.pyx
```diff
@@ -855,7 +883,15 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+        if desc == NULL:
+            desc = ""
+        desc = <FMIL3.fmi3_string_t>desc
```
modelonrobinandersson, line 889: And related to above, do we need to cast this again if we already did it on line 886?

src/pyfmi/fmi3.pyx
```diff
@@ -855,7 +883,15 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+
+        return pyfmi_util.decode(desc)
+
+    cdef _get_alias_description(self, FMIL3.fmi3_import_alias_variable_t* alias_variable):
+        cdef FMIL3.fmi3_string_t desc = <FMIL3.fmi3_string_t>FMIL3.fmi3_import_get_alias_variable_description(alias_variable)
         if desc == NULL:
             desc = ""
         desc = <FMIL3.fmi3_string_t>desc
```
modelonrobinandersson, line 897: Same comment here (i.e. `<FMIL3.fmi3_string_t>desc` twice)

src/pyfmi/fmi3.pyx
```diff
@@ -877,32 +913,47 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+            raise FMUException("The variable %s could not be found." % variable_name)
 
+    cdef _add_variable(self, FMIL3.fmi3_import_variable_t* variable):
         if variable == NULL:
             raise FMUException("Unknown variable. Please verify the correctness of the XML file and check the log.")
 
-        # TODO: Unnecessary to have alias_kind in FMI3?
-        # alias_kind = ?
         name        = pyfmi_util.decode(FMIL3.fmi3_import_get_variable_name(variable))
+        alias       = False
```
modelonrobinandersson, line 948: Is there 0 risk here that `variable` is not an alias? If so, feel free to resolve.

src/pyfmi/fmi3.pyx
```diff
@@ -1053,6 +1104,76 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+            raise FMUException("The variable %s could not be found."%variablename)
+
+        name = pyfmi_util.decode(FMIL3.fmi3_import_get_variable_name(variable))
+
+        return name
+
+    def get_variable_alias(self, variable_name):
+        """
+        Return a dict of all alias variables belonging to the provided variable
+        where the key are the names and the value indicati whether the variable
```
modelonrobinandersson, line 1135:
```suggestion
        where the key are the names and the value indicates if the variable
```
Also, if the value indicates if it is an alias or not, doesn't that conflict with the intention of the function since it is supposed to return all alias variables belonging to the provided variable? I.e. all `values` would be `True`?
Maybe I just misunderstood this function.

tests/test_fmi3.py
```diff
@@ -790,6 +790,50 @@ def test_generation_tool(self):
...
+    def test_get_model_variables_with_alias(self):
+        """Test get_model_variables including aliases."""
+        model_vars = self.fmu.get_model_variables(include_alias = True)
+        assert "v5" in model_vars
+        assert "v5_a1" in model_vars
+        assert "v5_a2" in model_vars
+
+        assert model_vars["v5"].description == "v5_desc"
+        assert model_vars["v5_a1"].description == ""
+        assert model_vars["v5_a2"].description == "v5_a2_desc"
```
modelonrobinandersson, line 809: Maybe also just add check that the class attribute `.alias` is True or False as expected.

modelonrobinandersson (commented):
Nice!

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: I'd say no for consistency with older FMI versions.

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: This does work; even simplified it a little more. I went with this since `desc = FMIL3. ... ` gave me conversion warnings.

However, I see that we have already had this in a couple of other places e.g., `get_model_version` etc., so this is a bit more uniform now.

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: No, fixed.

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: No risk; the FMIL types take care of this; here we have `fmi3_import_variable_t*`, aliases are `fmi3_import_alias_variable_t*`

tests/test_fmi3.py
PeterMeisrimelModelon: Added.

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: I suppose from a pure FMI3 perspective, this seems a little odd; but I think the final goal really is that an end-user can conveniently work with whatever alias of the eventual variable that they want and doesn't need to pay particular attention if it is the base or not. 

This particularly includes calling `fmu_get_variable_alias(<alias_name>)`

src/pyfmi/fmi3.pyx
modelonrobinandersson: I would argue that this is the time to change it if we want to. But since this is not "game breaking" I'm OK with it, feel free to resolve. I agree that consistency is good to maintain where possible.

src/pyfmi/fmi3.pyx
modelonrobinandersson: OK as long as it works for variables without any description I'm good! It was annoying previously when it crashed even though it looked identical to me between FMI 2 and 3.

src/pyfmi/fmi3.pyx
modelonrobinandersson: Nice!

modelonrobinandersson (approved):

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: The added tests specifically contains aliases without descriptions, so it appears to work.

PR #315, opened 2025-05-13
Adding getters and setters for the remaining int types + enum + boolean

PeterMeisrimelModelon:


tests/test_fmi3.py
```diff
@@ -580,35 +596,109 @@ def test_get_input_list(self):
...
         fmu_path = FMI3_REF_FMU_PATH / "Feedthrough.fmu"
         fmu = load_fmu(fmu_path)
-        int64_vals = fmu.get_int64([23, 24])
+        fmu.set(variable_name, value)
+        res = fmu.get(variable_name)
+        assert res.dtype == expected_dtype
+        assert res[0] == value
 
-        np.testing.assert_equal(np.array([0, 0]), int64_vals)
+    @pytest.mark.skipif(NUMPY_MAJOR_VERSION > 1, reason = "Error for numpy>=2")
```
PeterMeisrimelModelon, line 658: While the test is effectively somewhat duplicated, trying to follow the guideline of "don't use `ifs` depending on the parameterization in unit tests".

src/common/io.py
```diff
@@ -2603,6 +2603,10 @@ class ResultHandlerBinaryFile(ResultHandler):
     """
     Export an optimization or simulation result to file in Dymola's binary result file
     format (MATLAB v4 format).
+
+    All solution points are stored as float64s, this may lead to precision loss 
+    for very large values of (U)Int64 variables in FMI3.
+    Result storage of strings not supported.
```
modelonrobinandersson, line 2609:
```suggestion
    Result storage of strings is not supported.
```

src/pyfmi/fmi3.pyx
```diff
@@ -542,26 +563,651 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+            model.set_int64([234, 235],[2, 10])
+
+        Calls the low-level FMI function: fmi3SetInt64
+        """
+        cdef int status
+        cdef np.ndarray[FMIL3.fmi3_value_reference_t, ndim=1, mode='c'] input_valueref = np.asarray(valueref, dtype = np.uint32).ravel()
+        cdef np.ndarray[FMIL3.fmi3_int64_t, ndim=1, mode='c'] set_value = np.asarray(values, dtype = np.int64).ravel()
+
+        if np.size(input_valueref) != np.size(set_value):
+            raise FMUException('The length of valueref and values are inconsistent. Note: Array variables are not yet supported')
```
modelonrobinandersson, line 589: I think we should split this error check into a separate function since we have it now in so many places, and maybe we can also improve the error message to also include the size differences.

src/pyfmi/fmi3.pyx
```diff
@@ -542,26 +563,651 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+
+            valueref --
+                A list of value references.
+
+            values --
+                Values to be set.
+
+        Example::
+
+            model.set_enum([234, 235],["hello", "fmu"])
```
modelonrobinandersson, line 922:
```suggestion
            model.set_string([234, 235],["hello", "fmu"])
```

src/pyfmi/fmi3.pyx
```diff
@@ -542,26 +563,651 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+            np.size(input_valueref),
+            val,
+            np.size(values)
+        )
+        self._log_handler.capi_end_callback(self._max_log_size_msg_sent, self._current_log_size)
+
+        FMIL.free(val)
+
+        if status != 0:
+            raise FMUException('Failed to set the Enum values. See the log for possibly more information.')
```
modelonrobinandersson, line 950:
```suggestion
            raise FMUException('Failed to set the String values. See the log for possibly more information.')
```

src/pyfmi/fmi3.pyx
```diff
@@ -542,26 +563,651 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+
     def _get(self, variable_name):
         """
-        Helper method to get, see docstring on get.
+        Helper method to get, see docstring on get.
+        """
+        cdef FMIL3.fmi3_value_reference_t ref
+
```
modelonrobinandersson, line 995: Initially I was about to suggest that we change this to a dict and map each type to a function (similar to what we did in DumpDataFMI3), however, since this is a pyx file and the set/get-methods are invoked frequently I thought that maybe we get better performance as you have done it now, so let's leave it as that.
However, after a bit of reading I learnt that for regular "def" methods within a .pyx file, type annotations help optimize operations inside the method.

I would therefore propose at least we try to type annotate the get and set methods (if possible), and perhaps we might get a tiny bit of performance boost.

src/pyfmi/fmi3.pyx
```diff
@@ -542,26 +563,651 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+        elif basetype is FMI3_Type.UINT8:
+            return self.get_uint8([ref])
+        elif basetype is FMI3_Type.BOOL:
+            return self.get_boolean([ref])
+        elif basetype is FMI3_Type.STRING:
+            return self.get_string([ref])
+        elif basetype is FMI3_Type.ENUM:
+            return self.get_enum([ref])
+        else:
+            raise FMUException('Type not supported.')
```
modelonrobinandersson, line 1026:
```suggestion
            raise FMUException(f'Type '{basetype}' is not supported.')
```

tests/test_fmi3.py
```diff
@@ -580,35 +596,109 @@ def test_get_input_list(self):
...
+            ("get_int8", [11, 12], np.array([0, 0], dtype=np.int8), np.int8),
+            ("get_uint64", [25, 26], np.array([0, 0], dtype=np.uint64), np.uint64),
+            ("get_uint32", [21, 22], np.array([0, 0], dtype=np.uint32), np.uint32),
+            ("get_uint16", [17, 18], np.array([0, 0], dtype=np.uint16), np.uint16),
+            ("get_uint8", [13, 14], np.array([0, 0], dtype=np.uint8), np.uint8),
+            ("get_boolean", [27, 28], np.array([False, False], dtype=np.bool_), np.bool_),
+            ("get_enum", [32, 33], np.array([1, 1], dtype=np.int64), np.int64),
+        ]
+    )
+    def test_getX(self, function_name, valuerefs, expected_result, expected_dtype):
```
modelonrobinandersson, line 615: nice

modelonrobinandersson (commented):
Nice! Just very minor comments.

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: Split this out now, see changed.

As for the size check, I suggest we wait until we start on arrays and how exactly the inputs will look like then.
I'd imagine these would end up something like `fmu.set([1, 2], [[1, 2, 3], [[1, 2], [3, 4]]])` + then one could check against the actual sizes of the referenced variables.

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: I also tried a bit; I vaguely recall that Cython supposedly reduces something like this to `switch ... case` automatically, but I couldn't see any evidence of this in the generated C code. I think this largely goes back to the question if we should be using the Enums for the FMI3_Types in the first place, if Cython can recognize these as constants.

Not sure if there is more to annotate right now?

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: Also, I had considered if we could do something more along the lines of macros/inlining, but there would appear to be some issues with the class here, see comment under inlining here: https://cython.readthedocs.io/en/latest/src/userguide/pyrex_differences.html#cdef-inline

modelonrobinandersson: Nice. I agree with your comment (resolve).

src/pyfmi/fmi3.pyx
modelonrobinandersson: This is good for now. I think we can take most of it later with larger performance investigation when we are getting closer to finish this.

modelonrobinandersson (approved):

PR #320, opened 2025-05-28
Fixing issue with XML generation for dynamic_diagnostics

PeterMeisrimelModelon:


src/pyfmi/simulation/assimulo_interface_fmi2.pyx
```diff
@@ -279,48 +279,50 @@ cdef class FMIODE2(cExplicit_Problem):
...
-        else:
-            Jac = A
-            
-        if self._logging:
-            msg = preface + '</%s>'%(solver_info_tag)
-            self._model.append_log_message("Model", 4, msg)
+                Jac = A
+        finally:
+            # make sure to always generate valid XML, even if jacobian eval fails
+            if self._logging:
```
modelonrobinandersson, line 323: I just think we should add a boolean after line 291 such that we can verify we have successfully called `append_log_message` before we try to add closing tag below. I assume `self._update_model(t, y)` can fail or `self._f_nbr != 0` which would in turn not generate the log message, and then we end up here if anything fails and we are back to invalid XML. Proposal:

```
try:
  ...
  if ...
    append_log_message
    requires_closing_tag = True

  finally:
    if requires_closing_tag:
      append_log_message
```

modelonrobinandersson (approved):
I leave this as approved, if you add the boolean I dont think it needs another review. Well done!

src/pyfmi/simulation/assimulo_interface_fmi2.pyx
PeterMeisrimelModelon: Made some changes; please have another look.

Also learned this here:

```
def f():
    try:
        return 1
    finally:
        print("a")

a = f() # a = 1 & prints 
```

There doesn't seem to be a good way to test no + jacobian; there is something else blocking the jacobian call, don't think it is worth digging deeper for this here though.

src/pyfmi/simulation/assimulo_interface_fmi2.pyx
```diff
@@ -268,59 +268,59 @@ cdef class FMIODE2(cExplicit_Problem):
         """
         The jacobian function for an ODE problem.
         """
+        _log_requires_closing_tag = False
```
modelonrobinandersson: Since this is just a variable inside a function I dont see why it has to be named with leading `_`, no one will be able to access this anyway? But maybe I am missing something fundamental.

modelonrobinandersson (approved):

src/pyfmi/simulation/assimulo_interface_fmi2.pyx
modelonrobinandersson: Thats nice I didnt know about that either, thanks for sharing!

src/pyfmi/simulation/assimulo_interface_fmi2.pyx
PeterMeisrimelModelon: You are right, no point. Fixed.

PR #329, opened 2025-06-25
Fixing Type annotations for compatibility with Python < 3.10

PeterMeisrimelModelon:


tests/test_stream.py
```diff
@@ -107,7 +107,6 @@ def test_textiowrapper(self, fmu_path, test_class):
 @pytest.mark.assimulo
 class TestXML:
     """ Test other log related functions together with streams. """
-    @pytest.fixture(autouse=True)
```
PeterMeisrimelModelon, line 110: Caused issues with pytest on Python 3.9, redundant though.

.github/workflows/build.yml
```diff
@@ -8,12 +8,16 @@ permissions:
 jobs:
   linux:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python_version: [3.9.x, 3.11.x]
```
modelonrobinandersson, line 14: What does this do? Also it looks good to me, just wanted to learn.

PeterMeisrimelModelon: This is how one sets up for testing multiple versions of something.

Taken from https://docs.github.com/en/actions/use-cases-and-examples/building-and-testing/building-and-testing-python#using-multiple-python-versions + how we do it in Assimulo.

modelonrobinandersson: I see, thanks for sharing!

PR #332, opened 2025-07-02
Break out computations of calculated diagnostics variables into new utility class.

PeterMeisrimelModelon:
There are still too many hard-coded names to my liking, but I don't think we can fully get away from them either. That would require a larger effort and in particular moving the diagnostics related functionality into *.pyx, since the writing/reading would need to be aligned.

I think this is a step in the right direction either way.


src/common/diagnostics.py
```diff
@@ -36,6 +36,167 @@ class DiagnosticsBase:
...
+                index_calc = self._idx_map_calc_diags["nbr_state_limits"]
+                for i in range(self._number_states):
+                    if diag_data[index_diag_data + i] >= 1.0:
+                        ret[index_calc + i] = ret[index_calc + i] + 1
+
+        if update_cache:
+            self.calc_diags_vars_cache = ret
+        return ret
+    
+    @classmethod
```
PeterMeisrimelModelon, line 189: I would have liked these to be a vectorized variant of `get_calculated_diagnostics_point`, but this would become too much of a mess I fear. Plus we want `get_calculated_diagnostics_point` to be somewhat efficient.

So we simply need unit tests to make sure these do the same effective calculations.

src/common/io.py
```diff
@@ -184,27 +187,28 @@ def get_variable_index(self,name):
         # Strip name of spaces, for instance a[2, 1] to a[2,1]
         name = name.replace(" ", "")
 
-        try:
-            if isinstance(self, ResultDymolaBinary):
-                return self.name_lookup[encode(name)]
-            else:
-                return self.name_lookup[name]
-        except KeyError:
+        if isinstance(self, ResultDymolaBinary):
```
PeterMeisrimelModelon, line 190: Little bit of cleanup; we should try to avoid using Exceptions when possible.

src/common/diagnostics.py
```diff
@@ -36,6 +36,163 @@ class DiagnosticsBase:
...
+    diagnostics data produced from using 'dynamic_diagnostics'.
+    
+    This class contains functionality to explicitly store these on a per-point basis and
+    class-methods to perform the same calculations for trajectories."""
+    def __init__(self):
+        self._calc_diags_vars_cache: np.ndarray = np.array([])
+    
+    def setup_calculated_diagnostics_variables(self, diagnostics_vars: dict) -> dict:
+        """
+        Get calculated diagnostics variable names, start_values and descriptions.
```
modelonrobinandersson: Does this method setup and return diag related data?
I wonder if we should separate this into 2, or perhaps clarify the docstring or method name.

src/common/diagnostics.py
```diff
@@ -36,6 +36,163 @@ class DiagnosticsBase:
...
+    def _get_calc_diags_vars_cache(self) -> np.ndarray:
+        return self._calc_diags_vars_cache.copy()
+    def _update_calc_diags_vars_cache(self, val: np.ndarray) -> None:
+        self._calc_diags_vars_cache = val.copy()
+    calc_diags_vars_cache = property(
+        fget = _get_calc_diags_vars_cache, 
+        fset = _update_calc_diags_vars_cache
+    )
+
+    def get_calculated_diagnostics_point(self, diag_data: np.ndarray) -> np.ndarray:
```
modelonrobinandersson, line 146: Even here, I see "get_*" and I dont expect this function to run calculations based on the input. However, I am aware that this is because this MR is for the in fact calculated diagnostics data.
I don't have any good idea here yet but something to consider. Feel free to resolve this thread.

src/common/diagnostics.py
```diff
@@ -36,6 +36,163 @@ class DiagnosticsBase:
...
+                for i in range(self._number_states):
+                    if diag_data[index_diag_data + i] >= 1.0:
+                        ret[index_calc + i] = ret[index_calc + i] + 1
+
+        self.calc_diags_vars_cache = ret
+        return ret
+    
+    @classmethod
+    def get_cpu_time(cls, cpu_time_per_step: np.ndarray) -> np.ndarray:
+        """"Given cpu_time_per_step, return cumulative CPU time."""
```
modelonrobinandersson:
```suggestion
        """Given cpu_time_per_step, return cumulative CPU time."""
```

modelonrobinandersson: This function just returns the cumulative sum of the input. Does it really serve any purpose? Or should we rename it?

src/common/io.py
```diff
@@ -1672,57 +1654,56 @@ def _find_max_trajectory_length(self, trajectories):
...
+                    f"{DIAGNOSTICS_PREFIX}nbr_steps",
+                    ]
+            ):
+            return self._get_calculated_diagnostics_events_and_steps(name)
+        elif name.startswith(f"{DIAGNOSTICS_PREFIX}nbr_state_limits_step"):
+            return self._get_calculated_diagnostics_nbr_state_limits_step(name)
+        elif name == f"{DIAGNOSTICS_PREFIX}cpu_time":
+            return self._get_calculated_diagnostics_cpu_time()
+        else:
+            raise KeyError("Unknown calculated diagnostics variable requested.")
```
modelonrobinandersson:
```suggestion
            raise KeyError(f"Unknown calculated diagnostics variable of name '{name}' requested.")
```

tests/test_diagnostics.py
```diff
@@ -0,0 +1,219 @@
...
+FMI2_REF_FMU_PATH = Path(this_dir) / 'files' / 'reference_fmus' / '2.0'
+
+class ResultStoreCalcDiagnostics(ResultHandler):
+    """Result handler for testing explicit storage of calculated diagnostics."""
+    def __init__(self, model = None):
+        super().__init__(model)
+        self._diags_aux = DynamicDiagnosticsUtils()
+
+        self.supports['dynamic_diagnostics'] = True
+        self.diag_params : Union[dict, None] = None
```
modelonrobinandersson, line 39: If you are introducing `Union`, could you also update some of the other places where we use `a | b`? I think we did this due to older Python support (if I recall correctly `Union` is with 3.10 and newer). I think there are only a few of them.

tests/test_diagnostics.py
```diff
@@ -0,0 +1,219 @@
...
+            f"{DIAGNOSTICS_PREFIX}nbr_time_events":               np.array([0, 0, 0, 0, 1, 2, 2, 2, 3, 3]),
+            f"{DIAGNOSTICS_PREFIX}nbr_state_events":              np.array([0, 0, 1, 1, 1, 1, 2, 3, 3, 3]),
+            f"{DIAGNOSTICS_PREFIX}nbr_steps":                     np.array([1, 2, 2, 3, 3, 3, 3, 3, 3, 4]),
+            f"{DIAGNOSTICS_PREFIX}nbr_state_limits_step.state_x": np.array([0, 1, 1, 2, 2, 2, 2, 2, 2, 3]),
+        }
+
+        # verify via DynamicDiagnosticsUtils class functions on trajectories
+        res_trajectories = {}
+        res_trajectories[f"{DIAGNOSTICS_PREFIX}cpu_time"] = DynamicDiagnosticsUtils.get_cpu_time(
+            diagnostics_vars[f"{DIAGNOSTICS_PREFIX}cpu_time_per_step"])
```
modelonrobinandersson, line 132: With respect to my earlier comment, I guess I see now that the user is not aware of the underlying calculations to get from `*_per_step` to `cpu_time`. I wonder if the interface could be made differently though but I dont have any good suggestions right now, maybe this is something we could discuss separately.

tests/test_diagnostics.py
```diff
@@ -0,0 +1,219 @@
...
+        # verify via DynamicDiagnosticsUtils.get_calculated_diagnostics_point
+        dyn_diags_util = DynamicDiagnosticsUtils()
+        # set start values
+        calc_diags_points = {k: [v[0]] for k, v in dyn_diags_util.setup_calculated_diagnostics_variables(diagnostics_vars).items()}
+
+        n_points = len(diagnostics_vars[f"{DIAGNOSTICS_PREFIX}cpu_time_per_step"]) - 1 # -1 to adjust for start
+        # loop over points and build trajectories
+        for i in range(n_points):
+            idx = i + 1 # adjust for start
+            diags_input = np.array([v[idx] for v in diagnostics_vars.values()])
```
modelonrobinandersson:
```suggestion
        for i in range(1, n_points + 1):
            diags_input = np.array([v[i] for v in diagnostics_vars.values()])
```

tests/test_diagnostics.py
PeterMeisrimelModelon: There are no other instances left where we use this in PyFMI. 
Also, since we now have a GitHub action for Python 3.9, any issues would be highly visible.

src/common/diagnostics.py
PeterMeisrimelModelon: It is both setup + has a return ....

Effectively, the input is the output from `setup_diagnostics_variables` + we return not only variable names, but also start values and descriptions. By having an output, we cannot reduce this to just a constructor. 


This was split into 2 functions in the initial proof of concept, which got the specific feedback that it could/should be reduced to a single function.

An alternative would be that we make `diagnostics_vars` an input to the constructor instead. 

Improved the docstring a little.

src/common/diagnostics.py
PeterMeisrimelModelon: I think it serves the clear purpose that one can fully use this utility class to get all the calculated trajectories, even if some calculations, like this one, are quite simple. 

This keeps it simple and stupid; one should not need to know the associated calculations of the computed diagnostics variables to write a result handler that explicitly stores them.

tests/test_diagnostics.py
PeterMeisrimelModelon: See earlier comment.

src/common/diagnostics.py
PeterMeisrimelModelon: I'd like to think the docstring is appropriate, but function name could indeed be improved. 

Maybe simply `calculated_diagnostics_point`?

src/common/diagnostics.py
```diff
@@ -36,6 +36,180 @@ class DiagnosticsBase:
...
+class DynamicDiagnosticsUtils:
+    """Utility functionality for additional diagnostics variables calculated from 
+    diagnostics data produced from using 'dynamic_diagnostics'.
+    
+    This class contains functionality to explicitly compute and store these on a 
+    per-point basis and class-methods to perform the same calculations for trajectories.
+    
+    Minimal example for result handler using this class:
+
+    class ResultStoreCalcDiagnostics(ResultHandler):
```
PeterMeisrimelModelon, line 48: Also added this one

src/common/diagnostics.py
modelonrobinandersson: Could we call this method simply `prepare_diagnostics` or `prepare_calculated_diagnostics`?

src/common/diagnostics.py
```diff
@@ -36,6 +36,180 @@ class DiagnosticsBase:
...
+
+            dict: {calculated_diagnostics_name: (start_value, description)}
+        """
+        # Fixed variables
+        calc_diags = {
+            f"{DIAGNOSTICS_PREFIX}cpu_time"        : (0.0, "Cumulative CPU time"),
+            f"{DIAGNOSTICS_PREFIX}nbr_events"      : (0, "Cumulative number of events"),
+            f"{DIAGNOSTICS_PREFIX}nbr_time_events" : (0, "Cumulative number of time events"),
+            f"{DIAGNOSTICS_PREFIX}nbr_state_events": (0, "Cumulative number of state events"),
+            f"{DIAGNOSTICS_PREFIX}nbr_steps"       : (1, "Cumulative number of steps"), # XXX: is 1 as start actually correct?
```
modelonrobinandersson: Maybe try with 0 and see if it makes sense with a test. It would be strange with 1 as a start considering simulations could fail immediately?

src/common/diagnostics.py
```diff
@@ -36,6 +36,180 @@ class DiagnosticsBase:
...
+            "nbr_state_limits": len(calc_diags_names) - self._number_states,
+        }
+
+        idx_state_errors = None
+        for idx, key in enumerate(diagnostics_vars):
+            if key.startswith(f"{DIAGNOSTICS_PREFIX}state_errors."):
+                idx_state_errors = idx
+                break
+
+        diagnostics_vars_names = list(diagnostics_vars.keys())
```
modelonrobinandersson: Can we move this line up before the loop on line 92 `for v in diagnostic_vars.keys():` and then simply update the for loop to iterate on `diagnostics_vars_names`?

src/common/diagnostics.py
```diff
@@ -36,6 +36,180 @@ class DiagnosticsBase:
...
+            f"{DIAGNOSTICS_PREFIX}nbr_state_events": (0, "Cumulative number of state events"),
+            f"{DIAGNOSTICS_PREFIX}nbr_steps"       : (1, "Cumulative number of steps"), # XXX: is 1 as start actually correct?
+        }
+
+        # based on states
+        description = "Cumulative number of times '{}' limited the step-size."
+        start_value = 0
+        self._number_states = 0
+        for v in diagnostics_vars.keys():
+            prefix = f'{DIAGNOSTICS_PREFIX}state_errors.'
```
modelonrobinandersson: Let's move `prefix = ...` outside the loop.

src/common/diagnostics.py
```diff
@@ -36,6 +36,180 @@ class DiagnosticsBase:
...
+            "event_type": idx_event_type,
+            "state_errors": idx_state_errors, # index of first 'state_errors' variable in diag_vars
+        }
+
+        self.calc_diags_vars_cache = [v[0] for v in calc_diags.values()]
+        return calc_diags
+    
+    def _get_calc_diags_vars_cache(self) -> np.ndarray:
+        return self._calc_diags_vars_cache.copy()
+    def _set_calc_diags_vars_cache(self, val: np.ndarray) -> None:
```
modelonrobinandersson:
```suggestion
    def _get_calc_diags_vars_cache(self) -> np.ndarray:
        return self._calc_diags_vars_cache.copy()

    def _set_calc_diags_vars_cache(self, val: np.ndarray) -> None:
```

src/common/diagnostics.py
modelonrobinandersson: Yes lets go with that

tests/test_diagnostics.py
modelonrobinandersson: Perfect!

src/common/diagnostics.py
PeterMeisrimelModelon: Sure, going with `prepare_calculated_diagnostics` then.

src/common/diagnostics.py
PeterMeisrimelModelon: good catch!

src/common/diagnostics.py
PeterMeisrimelModelon: Sure, fixed.

src/common/diagnostics.py
PeterMeisrimelModelon: Actually just went ahead and fixed it - this was just wrong, see test that I added.

src/common/diagnostics.py
modelonrobinandersson: Sounds good! (resolve)

src/common/diagnostics.py
modelonrobinandersson: Looks good, glad that we got it fixed!

PR #333, opened 2025-07-07
Enable "dynamic_diagnostics" for FMI3

PeterMeisrimelModelon:


src/pyfmi/simulation/assimulo_interface_fmi3.pyx
```diff
@@ -455,10 +455,10 @@ cdef class FMIODE3(cExplicit_Problem):
                 solver.y = self._model.continuous_states.append(solver.y[-self._extra_f_nbr:])
             else:
                 solver.y = self._model.continuous_states
-
-        # # Get new nominal values.
-        if eInfo.nominalsOfContinuousStatesChanged: # TODO; this can overwrite explicit atol choices?
```
PeterMeisrimelModelon, line 460: See a few lines up

src/pyfmi/simulation/assimulo_interface_fmi3.pyx
```diff
@@ -455,10 +455,10 @@ cdef class FMIODE3(cExplicit_Problem):
...
-            solver.atol = 0.01*solver.rtol*self._model.nominal_continuous_states
+            # Get new nominal values.
+            if hasattr(solver, "atol"):
```
PeterMeisrimelModelon, line 459: Not true for e.g., ExplicitEuler

src/common/diagnostics.py
```diff
@@ -57,7 +58,7 @@ def setup_diagnostics_variables(model, start_time, options, solver_options):
                                     solver_name=="ExplicitEuler"
                                     )
 
-        states_list = model.get_states_list() if isinstance(model, FMUModelME2) else []
+        states_list = model.get_states_list() if isinstance(model, (FMUModelME2, FMUModelME3)) else []
```
modelonrobinandersson: Would it be possible to change this to `hasattr(model, get_states_list)`? With the purpose of making it more scalable and dont need updates int he future with additional classes.

src/pyfmi/simulation/assimulo_interface_fmi3.pyx
modelonrobinandersson: Good catch!

src/pyfmi/simulation/assimulo_interface_fmi3.pyx
```diff
@@ -505,7 +506,7 @@ cdef class FMIODE3(cExplicit_Problem):
                 for e in solver.get_weighted_local_errors():
                     diag_data[index] = e
                     index +=1
-            support_event_indicators = (solver_name=="CVode" or solver_name=="Radau5ODE" or solver_name=="LSODAR" or solver_name=="ImplicitEuler")
+            support_event_indicators = (solver_name=="CVode" or solver_name=="Radau5ODE" or solver_name=="ExplicitEuler" or solver_name=="LSODAR" or solver_name=="ImplicitEuler")
```
modelonrobinandersson, line 509: I saw this also in the other PR, I wonder if it would be better if we have a function or an enumeration or list and update this to `support_event_indicators = solver_supports_event_indicators(solver_name)`?
Ideally `solver.supports_event_indicators` but I think that would be out of scope.

src/common/diagnostics.py
PeterMeisrimelModelon: Added. 

In other places we have these quite a few of these hard-coded for performance reasons, but I don't think this is really relevant here; this is a one-time operation during setup.

src/pyfmi/simulation/assimulo_interface_fmi3.pyx
PeterMeisrimelModelon: I think the more important step is to cache this. But while we are add it, this should be based on capability flags, rather than hard-coded solver names.

There is a follow-up ticket for this; I updated it.

modelonrobinandersson: Sounds good to me, well done!

modelonrobinandersson (approved):

Issue #335, opened 2025-07-09
Issue with locating fmilib_shared on Linux


modelonrobinandersson:
Since we did some changes in `setup.py` a few releases ago it seems that users require the have `libfmilib_shared.so` available on the computer and set `LD_LIBRARY_PATH`.
```
In [1]: from pyfmi import load_fmu
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
Cell In[1], line 1
----> 1 from pyfmi import load_fmu

File /usr/../Python/pyfmi/__init__.py:24
     21 __all__ = ['fmi_algorithm_drivers', 'examples', 'fmi', 'common']
     23 #Import the model class allowing for users to type e.g.,: from pyfmi import FMUModelME1
---> 24 from pyfmi.fmi import load_fmu
     25 from pyfmi.fmi1 import FMUModelME1, FMUModelCS1
     26 from pyfmi.fmi2 import FMUModelME2, FMUModelCS2

ImportError: libfmilib_shared.so: cannot open shared object file: No such file or directory
```

Some local testing seems that it has something to do with https://github.com/modelon-community/PyFMI/commit/44b40aa01c69fed2e09ec515b9f2dd22fa284de5

This is not the behavior we want since we want users to be able to run PyFMI without having to care about separate dynamic libraries as default.

modelonrobinandersson:
@jschueller have you had any reports with issues with recent versions from conda?
I discovered this today when building locally

jschueller:
no such issues with conda since it packs all the dependencies where it should be

modelonrobinandersson:
@jschueller oh so conda will also install the dynamic library for FMIL as well when installing pyfmi? Does it install it within the PyFMI directory or together with other system libs?

jschueller:
yes I dont think it should be copied, instead rpath can be used, are you trying to do both ?

modelonrobinandersson:
@jschueller but if I bundle PyFMI as a wheel, that wheel will not work on its own and users would require users to get fmil as well separately right?

PR #336, opened 2025-07-10
Change how handling of dynamic lib of FMIL is done on Linux

modelonrobinandersson:
Potential fix for https://github.com/modelon-community/PyFMI/issues/335

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -111,6 +111,8 @@ static_link_gcc = "-static-libgcc"
 flag_32bit = "-m32"
 extra_c_flags = ""
 
+is_windows = sys.platform.startswith("win")
+
 # Fix path sep
 for x in sys.argv[1:]:
     if not x.find('--prefix'):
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -169,29 +175,37 @@
...
+
+        if not os.path.exists(path_to_dir):
+            continue
+
+        for file_name in os.listdir(path_to_dir):
+            full_path = os.path.join(path_to_dir, file_name)
+            if fmil_name in file_name and not file_name.endswith(".a"):
+                return shutil.copy2(full_path, os.path.join(".", "src", "pyfmi"))
+    raise Exception(
+        f"Could not find shared library '{fmil_name}' at either location:" + \
```
modelonrobinandersson:
```suggestion
        f"Could not find shared library '{fmil_name}' at any of the following locations:" + \
```

modelonrobinandersson:
@jschueller any inputs on these changes? Not able to assign you as a reviewer due to permission issues but any feedback here is much appreciated. I think I have gathered what we discussed in #335.

setup.py
```diff
@@ -119,7 +166,11 @@
         copy_args[copy_args.index(x)] = x.replace('/',os.sep)
     if not x.find('--fmil-home'):
         incdirs = [os.path.join(x[12:],'include')]
-        libdirs = [os.path.join(x[12:],'lib'), os.path.join(x[12:],'lib64')]
+        libdirs = [
+            os.path.join(x[12:],'lib'),
+            os.path.join(x[12:],'lib64'),
+            os.path.join(x[12:], 'lib', 'x86_64-linux-gnu'),
```
jschueller: I would use sysconfig.get_config_var('MULTIARCH') instead of hardcoding x86_64-linux-gnu

modelonrobinandersson: Fixed thanks!

setup.py
modelonrobinandersson: fixed

modelonrobinandersson:
> I can try to apply this on conda-forge to see if it still builds

thank you that would be great, let me know once you have an update.

setup.py
```diff
@@ -119,7 +167,11 @@
         copy_args[copy_args.index(x)] = x.replace('/',os.sep)
     if not x.find('--fmil-home'):
         incdirs = [os.path.join(x[12:],'include')]
-        libdirs = [os.path.join(x[12:],'lib'), os.path.join(x[12:],'lib64')]
+        libdirs = [
+            os.path.join(x[12:],'lib'),
+            os.path.join(x[12:],'lib64'),
+            os.path.join(x[12:], 'lib', sysconfig.get_config_var('MULTIARCH')),
```
jschueller: this does not work on windows because the MULTIARCH key is not available,
maybe we can add it to the list only if it exists:
```
if "MULTIARCH" in sysconfig.get_config_vars():
    libdirs.append(...
```

modelonrobinandersson: good catch, see https://github.com/modelon-community/PyFMI/pull/336/commits/9af1ad84fe1294d2c84078b46c6f25eff5cd0778

setup.py
```diff
@@ -111,15 +157,25 @@
...
     if not x.find('--fmil-home'):
         incdirs = [os.path.join(x[12:],'include')]
-        libdirs = [os.path.join(x[12:],'lib'), os.path.join(x[12:],'lib64')]
+        libdirs = [
+            os.path.join(x[12:],'lib'),
+            os.path.join(x[12:],'lib64'),
+        ]
+
+        multiarch = sysconfig.get_config_var('MULTIARCH')
+        if multiarch:
```
jschueller:
```suggestion
        if multiarch is not None:
```

jschueller (approved):

setup.py
```diff
@@ -169,29 +221,42 @@
...
-            raise Exception(
-                f"Could not find shared library 'fmilib_shared' at either location:" + \
-                    f"\n\t{', '.join(dirs_to_search)}")
-
-        if copy_gcc_lib:
-            path_gcc_lib = ctypes.util.find_library("libgcc_s_dw2-1.dll")
-            if path_gcc_lib is not None:
-                shutil.copy2(path_gcc_lib,os.path.join(".","src","pyfmi"))
-                gcc_lib = os.path.join(".","src","pyfmi","libgcc_s_dw2-1.dll")
+def find_dynamic_fmil_library():
```
PeterMeisrimelModelon: All for dividing things into smaller functions, but can we then also take e.g., `libdirs` and `bindirs` as explicit arguments?

setup.py
```diff
@@ -169,29 +221,42 @@
...
+            if fmil_name in file_name and not file_name.endswith(".a"):
+                return full_path
+
+    raise Exception(
+        f"Could not find shared library '{fmil_name}' at either locations:" + \
+            f"\n\t{', '.join(dirs_to_search)}")
+
+if 0 != sys.argv[1].find("clean"): # Dont check if we are cleaning!
+
+    use_dynamic_fmil_library = fmil_name.endswith("shared") # this should be improved in a future release
```
PeterMeisrimelModelon:
```suggestion
    use_dynamic_fmil_library = fmil_name.endswith("shared") # TODO: this should be improved in a future release
```
Without a TODO or similar, the chances of finding this are slim.

setup.py
```diff
@@ -290,6 +355,14 @@ def check_extensions():
                     include_path = incl_path,
                     compiler_directives={'language_level' : "3str"})
 
+    if not is_windows:
+        if use_dynamic_fmil_library:
+            if is_wheel_build:
+                extra_link_flags += ['-Wl,-rpath,$ORIGIN']
+            else:
+                extra_link_flags += [f'-Wl,-rpath,{os.path.dirname(fmil_shared)}']
```
PeterMeisrimelModelon:
```suggestion
    if not is_windows and use_dynamic_fmil_library:
        if is_wheel_build:
            extra_link_flags += ['-Wl,-rpath,$ORIGIN']
        else:
            extra_link_flags += [f'-Wl,-rpath,{os.path.dirname(fmil_shared)}']
```

PeterMeisrimelModelon (approved):

setup.py
modelonrobinandersson: Fixed

setup.py
modelonrobinandersson: Agree, fixed

setup.py
modelonrobinandersson: good catch, fixed as well

PR #337, opened 2025-07-16
Introducing base class for all result readers

PeterMeisrimelModelon:


src/common/io.py
```diff
@@ -146,31 +188,39 @@ def simulation_end(self):
...
         """
         self.options = options
 
-    def get_result(self):
+    def get_result(self) -> ResultReader:
         """
         Method for retrieving the result. This method should return a
-        result of an instance of ResultStorage or of an instance of a
-        subclass of ResultStorage.
+        result of an instance of ResultReader.
```
modelonrobinandersson, line 201: Maybe I am missing something but currently `class ResultHandler` is not extending from any other class but the docstring says it should return an instance of `ResultReader`.

modelonrobinandersson (commented):
Well done!

src/common/io.py
PeterMeisrimelModelon: `ResultHandler` is more of abstract base class, but for the actual implementations this is accurate, see e.g., `ResultHandlerBinaryFile.get_result` returns an `ResultDymolaBinary` instance.

Also the static type checking would find inconsistencies of function returns when inheriting.

modelonrobinandersson: Could we have ResultHandler also extend from abc.ABC just to make it more clear? Even though we dont have any abstract methods? I am not sure if its bad practice but just my initial thought.

PeterMeisrimelModelon: I also thought about it, but I'm not sure there is a point really. 

As you state, there are no real abstract methods, it is conceivable one just wants to do integrations points or diagnostics points. And without abstract methods, abstract base class has no meaning, doesn't even block instantiation of such classes.

modelonrobinandersson: OK sounds good, feel free to resolve.

PR #339, opened 2025-07-21
Adding get_capability_flags for FMI3 ME

PeterMeisrimelModelon:


src/pyfmi/fmil3_import.pxd
```diff
@@ -101,6 +101,41 @@ cdef extern from 'fmilib.h':
         fmi3_initial_enu_calculated = 3,
         fmi3_initial_enu_unknown    = 4
 
+    cdef enum fmi3_capabilities_enu_t:
+        fmi3_me_needsExecutionTool                     = 0,
+        fmi3_me_canBeInstantiatedOnlyOncePerProcess    = 1,
```
modelonrobinandersson, line 106: Is there any benefit in having three flags for the same flag? I.e. `fmi3_|me|cs|se|_canBeInstantiatedOnlyOncePerProcess`?

PeterMeisrimelModelon: Not really, other than it being the most simple&stupid solution to the fact that me|cs|se have different capability flags.
Either way, this is the way it is in FMILibrary, so no changing it anymore here, would otherwise be a breaking change.

PR #341, opened 2025-07-21
Adding get_derivatives_list

PeterMeisrimelModelon:


src/pyfmi/fmi3.pyx
```diff
@@ -1912,6 +1913,33 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+            An ordered dictionary with the derivative variables.
+        """
+        cdef FMIL3.fmi3_import_variable_list_t* variable_list
+        cdef FMIL.size_t                        variable_list_size
+        cdef FMIL3.fmi3_import_variable_t*      variable
+        variable_dict = {}
+
+        variable_list = FMIL3.fmi3_import_get_continuous_state_derivatives_list(self._fmu)
+        if variable_list == NULL:
+            raise FMUException("The returned state_derivatives list is NULL.")
```
modelonrobinandersson: I think we should make this error message more clear while we are here, considering this is what a user potentially sees when invoking `fmu.get_derivatives_list()`.

tests/test_fmi3.py
```diff
@@ -317,6 +317,33 @@ def test_get_states_list(self):
...
+
+        assert derx0.variability is FMI3_Variability.CONTINUOUS
+        assert derx1.variability is FMI3_Variability.CONTINUOUS
+
+        assert derx0.value_reference == 2
+        assert derx1.value_reference == 4
+
+        assert derx0.initial is FMI3_Initial.CALCULATED
+        assert derx1.initial is FMI3_Initial.CALCULATED
+
```
modelonrobinandersson, line 351: Let's just also verify we get the expected behavior for an FMU with no derivative (i.e. trigger the exception). Well done!

PeterMeisrimelModelon: Added, also for `get_states_list()`

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: Any suggestions? 

Practically speaking, this is static information from the `modelDescription.xml`, so I have a hard time seeing how this will fail, other than a bug in FMILibrary.

modelonrobinandersson: Ahh I see, I thought it was just that there are no state derivatives, but perhaps it is just an empty list then? Could we maybe simply write something in the line of 'Internal error when retrieving the state derivatives, please contact support.' or similar? I just thought that it could at least be good if we give a hint of what to do if a user would see that error.

tests/test_fmi3.py
modelonrobinandersson: Nice! (resolve)

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: No state derivatives is just an pointer to an empty list, but not NULL, see the tests added in https://github.com/modelon-community/PyFMI/pull/341/commits/1a672d863bd4eff789f24aa4f452f770783f6aff

How about: `"Unexpected failure in retrieving the continuous state derivatives list."` ? "Internal" doesn't have much meaning really, if you just use a package everything is internal. And I'd like to avoid "contact support" in Exception messages.

Realistically, it could fail with a malloc failure, but nothing else really (aside from bugs).

modelonrobinandersson: I see, I am good with your suggestion

PR #343, opened 2025-07-22
FMI3: get_output_list, get_output_dependencies(_kind), get_derivatives_dependencies(_kind)

PeterMeisrimelModelon:


src/pyfmi/fmi3.pyx
```diff
@@ -1941,6 +1973,201 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+        dependent on. Returns two dictionaries, one with the states and
+        one with the (continuous, float64) inputs.
+        """
+        # TODO: More than just float64 outputs?
+        # caching
+        if (self._outputs_states_dependencies is not None and
+            self._outputs_inputs_dependencies is not None):
+               return self._outputs_states_dependencies, self._outputs_inputs_dependencies
+
+        cdef FMIL3.fmi3_import_variable_t      *variable
```
PeterMeisrimelModelon, line 1988: This one is quite different from FMI2, since we changed the FMIL API to dependencies.

src/pyfmi/fmi3.pyx
```diff
@@ -1941,6 +1973,201 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
 
         return variable_dict
 
+    cpdef get_output_dependencies(self):
+        """
+        Retrieve the variables that the (continuous, float64) outputs are
+        dependent on. Returns two dictionaries, one with the states and
+        one with the (continuous, float64) inputs.
+        """
+        # TODO: More than just float64 outputs?
```
PeterMeisrimelModelon, line 1982: Follow-up; same for the other functions

src/pyfmi/fmi3.pyx
```diff
@@ -1941,6 +1973,201 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+            FMI3_DependencyKind.FIXED     (= 2)
+            FMI3_DependencyKind.TUNABLE   (= 3)
+            FMI3_DependencyKind.DISCRETE  (= 4)
+
+        """
+        # TODO: More than just float64 outputs?
+        self.get_output_dependencies()
+        return self._outputs_states_dependencies_kind, self._outputs_inputs_dependencies_kind
+
+    cpdef get_derivatives_dependencies(self):
```
PeterMeisrimelModelon, line 2078: Essentially copy paste of `get_output_dependencies()`, but with derivatives instead of outputs.

src/pyfmi/fmi3.pyx
```diff
@@ -1941,6 +1973,201 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+
+        return states, inputs
+
+    cpdef get_output_dependencies_kind(self):
+        """
+        Retrieve the 'kinds' that the (continuous, float64) outputs are
+        dependent on. Returns two dictionaries, one with the states
+        and one with the (continuous, float64) inputs. The list of 'kinds'::
+
+            FMI3_DependencyKind.DEPENDENT (= 0)
```
modelonrobinandersson: Should we omit listing the integer value? I was thinking with the following reasoning:
1. Purpose of enum is such that the user dont need to care about the integer value
2. If for some reason, the order would change (even though it likely never will) in the class, then the numerical value in this docstring is no longer true.
3. If users are in interactive shells looking at this enum, they will see both the name as well as the integer value

src/pyfmi/fmi3.pyx
```diff
@@ -1941,6 +1973,201 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+
+        return states, inputs
+
+    cpdef get_derivatives_dependencies_kind(self):
+        """
+        Retrieve the 'kinds' that the derivatives are
+        dependent on. Returns two dictionaries, one with the states
+        and one with the (continuous float64) inputs. The list of 'kinds'::
+
+            FMI3_DependencyKind.DEPENDENT (= 0)
```
modelonrobinandersson: Same here

src/pyfmi/fmi3.pyx
```diff
@@ -225,6 +232,16 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
 
+        # Caching
+        self._outputs_states_dependencies = None
+        self._outputs_inputs_dependencies = None
+        self._outputs_states_dependencies_kind = None
+        self._outputs_inputs_dependencies_kind = None
+        self._derivatives_states_dependencies = None
+        self._derivatives_inputs_dependencies = None
+        self._derivatives_states_dependencies_kind = None
+        self._derivatives_inputs_dependencies_kind = None
```
modelonrobinandersson, line 243: No action here needed now since this is how we also do it in FMI2, but I wonder if it would be better "implementation wise" if we would combine the `*dependencies` and `*dependencies_kind` into 1 sort of class instance or similar since they are so tied together. Just some food for thought, doesnt affect the user. Example
```
outputs_states_deps = MyClass(<something>) 
derivatives_inputs_deps = MyClass(<something else>)
...
```
Anyway, feel free to resolve.

PeterMeisrimelModelon: One could, there might be better ways to organize them. 
But then again, the current format is simple & stupid and thus quite good + consistent with FMI2. 

From a practical point of view; we only use the dependencies once for setup; so extra overhead is not too critical.

Resolving for now; can still change this later, if we want to.

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: Good point; removed them.

PR #351, opened 2025-08-22
Fixing caching issues with ResultDymolaBinary.get_variables_data

PeterMeisrimelModelon:


tests/test_io.py
```diff
@@ -1959,6 +1959,57 @@ def test_get_cpu_time_partial(self):
...
+
+        rdb = ResultDymolaBinary(opts["result_file_name"], allow_file_updates = True)
+        var_name_1 = "v"
+        var_name_2 = "h"
+
+        part_traj_v_1 = rdb.get_variables_data([var_name_1], start_index = 0, stop_index = 3)[0][var_name_1].x
+        part_traj_h_1 = rdb.get_variables_data([var_name_2], start_index = 0, stop_index = 3)[0][var_name_2].x
+
+        part_traj_v_2 = rdb.get_variables_data([var_name_1], start_index = 3, stop_index = 6)[0][var_name_1].x
+        part_traj_h_2 = rdb.get_variables_data([var_name_2], start_index = 3, stop_index = 6)[0][var_name_2].x
```
PeterMeisrimelModelon, line 1980: This exact line previously caused erroneous result, caching was only index based.

src/common/io.py
```diff
@@ -1505,10 +1506,9 @@ def _get_interpolated_trajectory(self, data_index: int, start_index: int = 0, st
         """ Returns an interpolated trajectory for variable of corresponding index 'data_index'. """
         self._verify_file_data()
 
-        index_in_cache = data_index in self._data_2
-        partial_cache_ok = self._can_use_partial_cache(start_index, stop_index)
-        if (index_in_cache and not self._allow_file_updates) or (index_in_cache and partial_cache_ok):
-            return self._data_2[data_index]
+        current_data = self._data_2.get(data_index, None)
+        if (current_data is not None) and self._can_use_partial_cache(current_data, stop_index):
+            return self._data_2[data_index][start_index:stop_index]
```
PeterMeisrimelModelon, line 1511: The slicing is new here, but looks correct to me. (Unless we fail to test some scenario)

src/common/io.py
```diff
@@ -1414,18 +1413,21 @@ def _get_name_dict(self):
...
 
-    def _can_use_partial_cache(self, start_index: int, stop_index: Union[int, None]):
-        """ Checks if start_index and stop_index are equal to the last cached indices. """
-        return self._allow_file_updates and (self._last_set_of_indices == (start_index, stop_index))
+    def _can_use_partial_cache(self, partial_trajectory: np.ndarray, stop_index: Union[int, None]) -> bool:
+        """Check if (partial) trajectory length is sufficent, or new data needs to be read."""
+        # no file updates = full read only
+        # otherwise check sufficient length is available
+        return (not self._allow_file_updates) or \
+                (self._allow_file_updates and (stop_index is not None) and (len(partial_trajectory) >= stop_index))
```
modelonrobinandersson:
```suggestion
        return not self._allow_file_updates or (stop_index is not None and len(partial_trajectory) >= stop_index)

```

src/common/io.py
```diff
@@ -1457,13 +1459,12 @@ def _get_trajectory(self, data_index, start_index = 0, stop_index = None):
...
-
-        index_in_cache = data_index in self._data_3
-        partial_cache_ok = self._can_use_partial_cache(start_index, stop_index)
-        if (index_in_cache and not self._allow_file_updates) or (index_in_cache and partial_cache_ok):
-            return self._data_3[data_index]
-        self._data_3[data_index] = self._read_trajectory_data(data_index, True, start_index, stop_index)[start_index:stop_index]
-        return self._data_3[data_index]
+        # caching
+        current_data = self._data_3.get(data_index, None)
+        if (current_data is not None) and self._can_use_partial_cache(current_data, stop_index):
```
modelonrobinandersson: Since this `if ...` is written now in several places I suggest we break it into a separate function, I mean only the expression being evaluated, that is:
```
if f(x):
   ...
```

modelonrobinandersson (commented):
Nice!

src/common/io.py
PeterMeisrimelModelon: Fixed; please have a look. 
Kept some extra parenthesis for readability though.

modelonrobinandersson (approved):

PR #352, opened 2025-08-22
Add get/set/serialize/deserialize/free/serialized_fmu_state_size

modelonrobinandersson:


src/pyfmi/fmi3.pyx
```diff
@@ -2681,6 +2698,266 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+        cdef FMIL3.fmi3_status_t status
+        cdef FMIL3.fmi3_FMU_state_t internal_state = state.fmu_state
+
+
+        if not self._supports_get_set_FMU_state():
+            raise FMUException('This FMU does not support get and set FMU-state')
+
+        if internal_state == NULL:
+            print("FMU-state does not seem to be allocated.")
+            return
```
modelonrobinandersson: This is the same in the FMI2-implementation, I dont know why there is a `print` followed by `return. Looks bad. I suggest changing it.

src/pyfmi/fmi3.pyx
```diff
@@ -2681,6 +2698,266 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+        cdef FMIL.size_t n_bytes
+        cdef np.ndarray[FMIL3.fmi3_byte_t, ndim=1, mode='c'] serialized_state
+
+        cap_me = FMIL3.fmi3_import_get_capability(self._fmu, FMIL3.fmi3_me_canSerializeFMUState)
+        cap_cs = FMIL3.fmi3_import_get_capability(self._fmu, FMIL3.fmi3_cs_canSerializeFMUState)
+        if not cap_me and not cap_cs:
+            raise FMUException('This FMU dos not support serialization of FMU-state')
+
+        n_bytes = self.serialized_fmu_state_size(state)
+        serialized_state = np.empty(n_bytes, dtype=np.uint8)
```
modelonrobinandersson: Just a note that `dtype=np.uint8` is different between FMI2/FMI3, that is because the C type `fmi2Byte` and `fmi3Byte` are defined differently (`char` vs `uint8`). I just mention it for knowledge sharing because I missed that myself.

src/pyfmi/fmil3_import.pxd
```diff
@@ -37,6 +39,7 @@ cdef extern from 'fmilib.h':
     ctypedef void*    fmi3_instance_environment_t
     ctypedef char*    fmi3_string_t
     ctypedef bool     fmi3_boolean_t
+    ctypedef uint8_t  fmi3_byte_t
```
modelonrobinandersson, line 42: Different between fmi2/fmi3.

tests/test_fmi3_states.py
```diff
@@ -0,0 +1,211 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+from pathlib import Path
```
modelonrobinandersson, line 18: Just some comments here:
1. Some tests might like odd/random, but I wrote them that way because we discovered many bugs in the FMI2-implementation when we tried those workflows.
2. I have added _some_ of the reference FMUs for testing. Feel free to add more but I thought it was sufficient to at least have some of them and not all.
3. Note that I left `test_todo`, for me that fails locally but I think that kind of workflow should work without errors.

src/pyfmi/fmi3.pyx
```diff
@@ -2681,6 +2698,263 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+        cdef FMIL3.fmi3_status_t status
+        cdef FMIL3.bool cap_me, cap_cs
+
+        cdef FMIL.size_t n_bytes
+        cdef np.ndarray[FMIL3.fmi3_byte_t, ndim = 1, mode = 'c'] serialized_state
+
+        cap_me = bool(FMIL3.fmi3_import_get_capability(self._fmu, FMIL3.fmi3_me_canSerializeFMUState))
+        cap_cs = bool(FMIL3.fmi3_import_get_capability(self._fmu, FMIL3.fmi3_cs_canSerializeFMUState))
+        if not cap_me and not cap_cs:
+            raise FMUException('This FMU dos not support serialization of FMU-state')
```
chria:
```suggestion
            raise FMUException('This FMU does not support serialization of FMU-state')
```

PR #355, opened 2025-09-04
Fixing issue with interpolation to diagnostics time if there is only …

PeterMeisrimelModelon:
…a single time point; release prep


src/common/io.py
```diff
@@ -1523,11 +1523,15 @@ def _get_interpolated_trajectory(self, data_index: int, start_index: int = 0, st
         time_vector      = self._read_trajectory_data(0, False, start_index, stop_index)
         data             = self._read_trajectory_data(data_index, False, start_index, stop_index)
 
-        n = min(len(time_vector), len(data)) # safety for reading wiith allow_file_updates = True
-        f = scipy.interpolate.interp1d(time_vector[:n], data[:n], fill_value = "extrapolate")
+        n = min(len(time_vector), len(data)) # safety for reading with allow_file_updates = True
+        if n == 1: # length 1 would lead to issues in interpolation
```
PeterMeisrimelModelon, line 1527: Seemed easiest to catch it just before the interpolation call.

PR #356, opened 2025-09-09
Fixing bug with result storing of bool and enums

PeterMeisrimelModelon:
This came in with the refactoring in https://github.com/modelon-community/PyFMI/pull/309 and was not caught by regressions tests.


tests/conftest.py
```diff
@@ -48,8 +48,30 @@ def download_url(url, save_file_to):
 
             with ZipFile(zip_path, 'r') as zf:
                 for fobj in zf.filelist:
-                    # For now, only unpack FMI 3.0 FMUs
-                    if (fobj.filename.startswith('3.0') or fobj.filename.startswith('2.0')) and fobj.filename.endswith('.fmu'):
+                    if fobj.filename.startswith('3.0') and fobj.filename.endswith('.fmu'):
+                        zf.extract(fobj, zip_unzip_to)
+                        with open(md5_file, 'w') as f:
+                            f.write(md5)
+    # NOTE: FMI2 Feedthrough.fmu is broken in v0.0.39; use v0.0.38 for FMI2 instead
```
PeterMeisrimelModelon, line 55: Note: https://github.com/modelica/Reference-FMUs/pull/656

src/common/io.py
```diff
@@ -2896,8 +2896,8 @@ def _get_sorted_vars(self):
             data_types = [
                 fmi.FMI_REAL,
                 fmi.FMI_INTEGER,
+                fmi.FMI_ENUMERATION,
```
PeterMeisrimelModelon, line 2899: in fmi_util, we loop over `real`s, then `integer`s, then `bool`s. In FMI2 `enum`s are `int`s and thus need this ordering to be correctly bundled. 

Not super pretty, but we are not expecting to touch this again.

MikaelJohnssonModelon: ok

PR #358, opened 2025-09-12
Some cleanup in master alg

PeterMeisrimelModelon:
The back and forth reshaping/flattening seems redundant and actually causes numpy DeprecationWarnings in other incoming tests. Note: `reshape((-1, 1))` converts an array of shape `(N,)` to an array of shape `(N, 1)`, i.e. simple 1D array to 2D column array.

Broke out the fix to a separate MR for ease of review.


filip-stenstrom (approved):
Looks OK from what I can tell

PR #359, opened 2025-09-15
Co-sim master; support step-size downsampling

PeterMeisrimelModelon:
Could optionally also add some interaction between `result_downsampling_factor` and `step_size_downsampling_factor`, e.g., 

1. `result_downsampling_factor` default values given by `step_size_downsampling_factor`
2. Enforce (or warn if not) `result_downsampling_factor[model]` be an integer multiple of `step_size_downsampling_factor[model]`.

### DIFF
--- a/src/pyfmi/master.pyx
+++ b/src/pyfmi/master.pyx
@@ -165,8 +165,8 @@ cdef finalize_result_objects(object models_dict):
         
 def init_f(y, master):
     u = master.L.dot(y)
-    master.set_connection_inputs(u)
-    return y - master.get_connection_outputs()
+    master.set_connection_inputs(u, initialize = True)
+    return y - master.get_connection_outputs(initialize = True)
     
 def init_f_block(ylocal, master, block):
     y = np.zeros((master._len_outputs))
@@ -186,7 +186,7 @@ def init_f_block(ylocal, master, block):
     
 def init_jac(y, master):
     u = master.L.dot(y)
-    master.set_connection_inputs(u)
+    master.set_connection_inputs(u, initialize = True)
     D = master.compute_global_D()
     DL = D.dot(master.L)
     return np.eye(*DL.shape) - DL
@@ -198,7 +198,7 @@ class MasterAlgOptions(OptionBase):
     Options::
 
         step_size --
-            Specfies the global step-size to be used for simulating
+            Specifies the global step-size to be used for simulating
             the coupled system.
             Default: 0.01
             
@@ -344,6 +344,14 @@ class MasterAlgOptions(OptionBase):
             Usage with 'error_controlled' = True is not supported.
             Example: If set to 2: Result contains only every other communication point.
             Default: 1 (no downsampling)
+        # TODO: Interaction with step_size_downsampling_factor?
+
+        step_size_downsampling_factor -- 
+            Dictionary {model: int > 0}. 
+            A given model only updates updates its in&outputs every
+            <step_size_downsampling_factor>-th communication point.
+            Usage with 'error_controlled' = True is not supported.
+            Default: {m: 1 for m in models}
     """
     def __init__(self, master, *args, **kw):
         _defaults= {
@@ -373,7 +381,9 @@ class MasterAlgOptions(OptionBase):
         "experimental_output_solve":False,
         "force_finite_difference_outputs": False,
         "num_threads":None,
-        'result_downsampling_factor': 1
+        "result_downsampling_factor": 1,
+        # TODO: result_downsampling_factor per model? Simple number = all models ? 
+        "step_size_downsampling_factor" : dict((model, 1) for model in master.models),
         }
         super(MasterAlgOptions,self).__init__(_defaults)
         self._update_keep_dict_defaults(*args, **kw)
@@ -384,7 +394,7 @@ cdef class Master:
     cdef public object opts
     cdef public object models_dict, L, L_discrete
     cdef public object _ident_matrix
-    cdef public object y_prev, yd_prev, input_traj
+    cdef public object y_prev, yd_prev, input_traj, y_discrete_prev
     cdef public object DL_prev
     cdef public int algebraic_loops, storing_fmu_state
     cdef public int error_controlled, linear_correction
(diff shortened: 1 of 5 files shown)


src/pyfmi/master.pyx
```diff
@@ -403,9 +440,10 @@ cdef class Master:
     cdef public int _display_counter
     cdef public object _display_progress
     cdef public double _time_integration_start
-    cdef public int result_downsampling_factor
-    cdef public int _step_number
+    cdef public dict result_downsampling_factor
+    cdef public long long _step_number
```
PeterMeisrimelModelon, line 440: Safety; so we don't run into issues with very long simulations.

src/pyfmi/master.pyx
```diff
@@ -373,18 +386,42 @@ class MasterAlgOptions(OptionBase):
...
+                result_downsampling_factor = a.pop("result_downsampling_factor")
+        if result_downsampling_factor is not None:
+            self.__setitem__("result_downsampling_factor", result_downsampling_factor)
+
         self._update_keep_dict_defaults(*args, **kw)
 
+    def __setitem__(self, key, value):
+        # OptionBase enforce same type (e.g., int, dict) for inputs; exceptions to this format here
+        if key == "result_downsampling_factor" and not isinstance(value, dict):
+            warnings.warn(
```
PeterMeisrimelModelon: Also, multi-type option values, e.g., dict. vs int. are a pain when integrated into other tools. Proposing to keep it simple by only allowing `dicts` in the future.

src/pyfmi/master.pyx
```diff
@@ -1581,8 +1681,11 @@ cdef class Master:
         print('')
         print('Simulation interval      : ' + str(start_time) + ' - ' + str(final_time) + ' seconds.')
         print('Elapsed simulation time  : ' + str(time_stop-time_start) + ' seconds.')
-        for model in self.models:
-            print(' %f seconds spent in %s.'%(self.elapsed_time[model],model.get_name()))
+        if self.opts["execution"] == "parallel":
```
PeterMeisrimelModelon, line 1680: Would currently just show 0, since we do not do time measurements when running in parallel, likely due to GIL issues.

tests/test_fmi_master.py
```diff
@@ -396,155 +409,412 @@ def do_step2(current_t, step_size, new_step=True):
...
-        for f in [2, 3, 4, 5, 10, 100, 250, 600, 3000]:
-            opts_update = {'result_downsampling_factor': f}
-            models, connections = self._load_basic_simulation()
-            test_res = self._sim_basic_simulation(models, connections, opts_update)
+        np.testing.assert_array_equal(res[0]["Float64_discrete_output"], [0, 0, 0, 0, 0, 0])
+        np.testing.assert_array_equal(res[1]["Float64_discrete_output"], [0, 2, 4, 6, 8, 10])
 
-            test_res_trajs = (test_res[models[0]]['x1'], test_res[models[1]]['x2'])
-            self._verify_downsample_result(ref_res_trajs, test_res_trajs, 2000, f)
+class Test_Master_Result_Downsampling:
```
PeterMeisrimelModelon, line 452: Split these tests out into a separate class.
Also, changed them to use Reference FMUs + tuned them a bit to run fast and be more explicit.

src/pyfmi/master.pyx
```diff
@@ -337,13 +336,27 @@ class MasterAlgOptions(OptionBase):
             Default: False
 
         result_downsampling_factor --
-            int > 0, only save solution to result every
```
chria, line 340: Why not still allow a "global" downsampling factor?

src/pyfmi/master.pyx
```diff
@@ -337,13 +336,27 @@ class MasterAlgOptions(OptionBase):
...
             Usage with 'error_controlled' = True is not supported.
-            Example: If set to 2: Result contains only every other communication point.
-            Default: 1 (no downsampling)
+            It is not required to have all models as keys,
+            missing ones take the default value of 1.
+            Default: {m: 1 for m in models} (no downsampling)
+
+        step_size_downsampling_factor -- 
+            Dictionary {model: int > 0}. 
+            A given model only updates updates its in&outputs every
```
chria:
```suggestion
            A given model only updates its inputs and outputs every
```

src/pyfmi/master.pyx
```diff
@@ -337,13 +336,27 @@ class MasterAlgOptions(OptionBase):
...
+            <step_size_downsampling_factor>-th communication point.
+            Usage the following options values is not supported:
```
chria:
```suggestion
            The following options values is not supported together with the step-size downsampling option:
```

src/pyfmi/master.pyx
chria: How do you mean with integration into other tools?

src/pyfmi/master.pyx
```diff
@@ -831,21 +895,21 @@ cdef class Master:
...
+        cdef int i, inext, status = 0
+        cdef np.ndarray[FMIL2.fmi2_real_t, ndim=1, mode='c'] yd    = np.empty((self._len_outputs))
+        cdef np.ndarray[FMIL2.fmi2_real_t, ndim=1, mode='c'] ydtmp = np.empty((self._len_outputs))
         cdef np.ndarray y_last = None
         
         if self.opts["extrapolation_order"] > 0:
             if self.max_output_derivative_order > 0 and not self.opts["force_finite_difference_outputs"]:
                 for model in self.models_dict.keys():
-                    #yd.extend(model.get_output_derivatives(self.models_dict[model]["local_output"], 1))
+                    # Note: _downsampling_skip if supporting step_size_downsampling_factor + extrapolation_order > 0
```
chria: Not sure that I understand this comment? Can you elaborate?

src/pyfmi/master.pyx
PeterMeisrimelModelon: Right, forgot to add this as a discussion point.

I do see the point of convenience here, but at the same time trying to avoid option values of fundamental different types, if possible.

src/pyfmi/master.pyx
PeterMeisrimelModelon: Changed to `# Note: Will require _downsampling_skip if supporting step_size_downsampling_factor ...`

These are simply reminders/TODOs about various places in the code where we will need to make changes/additions, if we added support for `step_size_downsampling_factor` + a given option combination, here, e.g., `extrapolation_order > 0`.

PR #360, opened 2025-09-22
Improved API for log category interaction

PeterMeisrimelModelon:


src/pyfmi/fmi2.pyx
```diff
@@ -1378,42 +1378,40 @@ cdef class FMUModelBase2(FMI_BASE.ModelBase):
...
+        cdef FMIL2.fmi2_boolean_t log
         cdef int                  status
-        cdef FMIL.size_t          n_cat = len(categories)
-        cdef list valid_categories
+        cdef FMIL.size_t          n_cat = np.size(categories)
+        cdef FMIL2.fmi2_string_t* val
 
         if logging_on:
-            self.callbacks.log_level = FMIL.jm_log_level_warning
+            self.set_log_level(7) # FMIL.jm_log_level_all
```
PeterMeisrimelModelon: This is changed behavior and `ALL = 7` might be overkill in many situations, however, warning is almost certainly wrong in most cases.

In practice, this likely needs to be combined with a suitable `set_log_level()` call afterwards.

src/pyfmi/fmi2.pyx
```diff
@@ -1378,42 +1378,40 @@ cdef class FMUModelBase2(FMI_BASE.ModelBase):
...
             log = 0
 
         self._enable_logging = bool(log)
 
-        if n_cat > 0:
-            valid_categories = self.get_categories()
-
-        cdef FMIL2.fmi2_string_t* val = <FMIL2.fmi2_string_t*>FMIL.malloc(sizeof(FMIL2.fmi2_string_t)*n_cat)
-        for i, c in enumerate(categories):
-            if c not in valid_categories:
```
PeterMeisrimelModelon, line 1413: Removing the error check, since having log categories not visible in the `modelDescription.xml` is a valid use.

src/pyfmi/fmi2.pyx
```diff
@@ -1424,22 +1422,36 @@ cdef class FMUModelBase2(FMI_BASE.ModelBase):
         if status != 0:
             raise FMUException('Failed to set the debugging option.')
 
-    def get_categories(self):
+    def get_categories(self) -> list[str]:
         """
-        Method used to retrieve the logging categories.
+        [DEPRECATED] Method used to retrieve the logging categories.
```
PeterMeisrimelModelon: deprecated only means it will be discontinued with FMI3; no intention to actually remove this, hence no warning.

src/pyfmi/fmi2.pyx
```diff
@@ -1424,22 +1422,36 @@ cdef class FMUModelBase2(FMI_BASE.ModelBase):
...
+        Use 'get_log_categories' to get both categories and descriptions instead
 
         Returns::
 
             A list with the categories available for logging.
         """
+        return list(self.get_log_categories().keys())
+
+    def get_log_categories(self) -> dict[str, str]:
```
PeterMeisrimelModelon, line 1436: Open to other ideas to how to design this API, this variant seems very sensible to me though.

src/pyfmi/fmi3.pyx
```diff
@@ -2533,6 +2533,73 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
         else:
             raise FMUException('Logging is not enabled')
 
+    def set_debug_logging(self, logging_on, categories = []):
```
PeterMeisrimelModelon, line 2536: Essentially copy-paste from FMI2; minus `get_categories`.

PeterMeisrimelModelon (commented):
Note: There is also log category related functionality for `CoupledFMUModelME2`, but that one is in a rather questionable state to begin with, since the getting/setting of categories is not done on a per-model basis??

Could just skip that one for now and get it right with FMI3.

src/pyfmi/fmi2.pyx
chria: Is this used somewhere else which expects it to be set to warning? Can this have a large performance regressions somewhere?

src/pyfmi/fmi2.pyx
chria: Is it according to spec?

src/pyfmi/fmi2.pyx
chria: I don't think that we should add a deprecation here. In FMI2 this will not go away and we will not remove it here. For FMI3 - things are expected to change but no need to "deprecate things here".

src/pyfmi/fmi2.pyx
chria: Another approach would be to leave it as is and add a method for just getting the description. Similar to how we do it for variables. I.e.:

`categories = get_categories()
category_desc = get_category_description(categories[0])`

src/pyfmi/fmi2.pyx
PeterMeisrimelModelon: Yes. Spec says:

`Note that since element <LogCategories> is optional, an FMU does not need to expose its log categories.]`

src/pyfmi/fmi2.pyx
PeterMeisrimelModelon: That is an option, no very strong opinions here. 

My thinking here is primarily that:

1. `get_categories` is a bad name to begin with, categories of ...what?
2. Getting both names and descriptions jointly is sensible, for e.g., GUI visualization.
3. If descriptions are not needed, just use the keys. Also lends itself nicely to the fact that default iterations on dictionaries are on the keys.

src/pyfmi/fmi2.pyx
PeterMeisrimelModelon: I looked into it a bit more, this does affect e.g., logging of FMIL CAPI calls.

I'm leaning towards maybe simply leaving this part as is for FMI2, to not cause any regressions.

For FMI3 and forward, I suggesting we fully remove this implicit setting of the `log_level`, since it manipulates the `log_level` specified when loading the FMU. This would make `set_debug_logging` a pure function to toggle certain log categories ON/OFF, effectively a direct forward to the `fmi3SetDebugLogging` CAPI call (making the name justified), no other side-effects.

Thoughts?

src/pyfmi/fmi2.pyx
PeterMeisrimelModelon: Removed the [DEPRECATED].

src/pyfmi/fmi2.pyx
chria: I agree with you on leaving as is for FMI2.

For FMI3 - what does it mean to toggle a log category on/off? How does it related to the log level? Do you have to set both to actually get the message? And what log level would a specific category require?

src/pyfmi/fmi2.pyx
chria: Ok, lets add this method.

src/pyfmi/fmi2.pyx
PeterMeisrimelModelon: Actually, after some more cross-referencing with other parts, I created an issue on the standard: https://github.com/modelica/fmi-standard/issues/2060

But I think this error check is not required.

src/pyfmi/fmi2.pyx
PeterMeisrimelModelon: To provide the full picture:
FMI "only" has the **log statuses**: Ok, warning, discard, error, fatal
In FMIL we have more **log levels** (7), debug, verbose, info, warning, error, fatal, nothing (0). These are the ones set when loading an FMU, or directly via `fmu.set_log_level()`.

Mappings are 
- FMI OK & discard = FMIL.INFO
- FMI warning = FMIL.warning
- FMI error = FMIL.error
- FMI fatal = FMIL.fatal

During FMIL log forwarding, we discard any messages less severe than the given FMIL log level, e.g., with **FMIL.warning(3)**, messages to `logMessage` with `FMI_Status = fmiOK` are **not** logged.

For all practical purposes, FMIL log level= info will capture all messages coming from the FMU, higher log level will only yield logging of CAPI callbacks from FMIL.

So, for practical logging, you need both the correct logging categories toggled on in the FMU + sufficiently high log level set via `fmu.set_log_level()`. 

However, `set_debug_logging` is used to toggle on/off a given list of categories, so it should be feasible to first toggle on ALL and then disable a few. 

I think implicit adjustments to a lower FMIL log levels are an easy trap here. Toggling off log categories does NOT mean there will be no more logging.

A point could be made that by turning ON log categories, the log level should be adjusted upwards to at least INFO.

chria: Thanks for the input. So what does this leave us? We keep FMI2 as is and then do some adjustments for FMI3?

PeterMeisrimelModelon: Proposal:
- FMI2 leave as is.
- FMI3, skip `set_log_level()` adjustments in `set_debug_logging()`, but more explicit about need appropriate `set_log_level()` in the docstring.

chria: Ok sounds good.

PR #364, opened 2025-10-08
Automatically disable logging on the FMU level when exceeding the max…

PeterMeisrimelModelon:
…imal log size

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -11,8 +11,6 @@
     * Fixed an issue where `FMUModel(CS|ME)1.get_author()` returned a byte string, rather than a normal string.
     * Logging will (by default) be explicitly be disabled by calling `fmiSetDebugLogging` 
       and setting `log_level` to `0` when exceeding `set_max_log_size`.
-      * Use `fmu.set_disable_fmu_logging_after_max_log_size(<bool>)` to toggle this behavior.
-      * Use `fmu.get_disable_fmu_logging_after_max_log_size()` to query behavior, `default = True`.
 
 --- PyFMI-2.18.3 ---
     * Fixed a bug introduced in PyFMI 2.18.0 causing incorrect result storing for `boolean` and `enum` variables
--- a/src/pyfmi/fmi_base.pyx
+++ b/src/pyfmi/fmi_base.pyx
@@ -78,7 +78,6 @@ cdef class ModelBase:
         self._invoked_dealloc = 0 # Set to 1 when __dealloc__ is called
         self._result_file = None
         self._log_handler = LogHandlerDefault(self._max_log_size)
-        self._disable_logging_on_max_log_size = 1
 
     def _set_log_stream(self, stream):
         """ Function that sets the class property 'log_stream' and does error handling. """
@@ -399,7 +398,7 @@ cdef class ModelBase:
         return "Model"
 
     def _deactivate_logging(self):
-        pass # to be implemented in class inheriting from ModelBase
+        raise NotImplementedError
         
     def extract_xml_log(self, file_name=None):
         """
@@ -531,8 +530,7 @@ cdef class ModelBase:
     cdef inline str _increment_log_size_and_check_max_size(self, str msg):
         if self._current_log_size + len(msg) > self._max_log_size:
             msg = "The log file has reached its maximum size and further log messages will not be saved. To change the maximum size of the file, please use the 'set_max_log_size' method.\n"
-            if self._disable_logging_on_max_log_size:
-                self._deactivate_logging()
+            self._deactivate_logging()
             self._max_log_size_msg_sent = True
         self._current_log_size = self._current_log_size + len(msg)
         return msg
@@ -701,26 +699,6 @@ cdef class ModelBase:
         else:
             raise NotImplementedError(f"Unable to convert filter: {expression=}")
         return regexp
-
-    def set_disable_fmu_logging_after_max_log_size(self, val: bool):
-        """
-        Set if logging from the FMU will be disabled via calling
-            `fmu.set_log_level(0)`
-            `fmu.set_debug_logging(False)`
-        upon reaching max_log_size.
-
-        Set this to `False` for continued logging using `additional_logger`.
-        """
-        self._disable_logging_on_max_log_size = int(val)
-
-    def get_disable_fmu_logging_after_max_log_size(self) -> bool:
-        """
-        Returns boolean if logging from the FMU will be disabled via calling
-            `fmu.set_log_level(0)`
-            `fmu.set_debug_logging(False)`
-        upon reaching max_log_size.
-        """
-        return bool(self._disable_logging_on_max_log_size)
 
 class PyEventInfo(): # TODO: Should this be a cpdef + FMIX variants?
     pass
--- a/tests/test_log.py
+++ b/tests/test_log.py
@@ -31,22 +31,10 @@ import numpy as np
 file_path = os.path.dirname(os.path.abspath(__file__))
 logs = os.path.join(file_path, "files", "Logs")
 
-this_dir = Path(__file__).parent.absolute()
+this_dir = Path(__file__).parent
 FMI1_REF_FMU_PATH = Path(this_dir) / 'files' / 'reference_fmus' / '1.0'
 FMI2_REF_FMU_PATH = Path(this_dir) / 'files' / 'reference_fmus' / '2.0'
 FMI3_REF_FMU_PATH = Path(this_dir) / 'files' / 'reference_fmus' / '3.0'
-
-def _get_fmu_class_logging_deactivation(parent_class):
-    class FMUModelLoggingDeactivation(parent_class):
-        """Testing class that injects some logging and test 
-        that `deactivate_logging` was called."""
-        def __init__(self, *args, **kwargs):
-            super().__init__(*args, **kwargs)
-            self.deactivate_called = False
-        def _deactivate_logging(self):
-            self.deactivate_called = True
-            return super()._deactivate_logging()
-    return FMUModelLoggingDeactivation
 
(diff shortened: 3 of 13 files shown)


src/pyfmi/fmi1.pyx
```diff
@@ -780,6 +780,10 @@ cdef class FMUModelBase(FMI_BASE.ModelBase):
         if status != 0:
             raise FMUException('Failed to set the String values. See the log for possibly more information.')
 
+    def _deactivate_logging(self):
```
PeterMeisrimelModelon, line 783: Split this out into the different FMI versions since the CAPI call differs based on FMI version.

src/pyfmi/fmi_base.pyx
```diff
@@ -697,6 +702,26 @@ cdef class ModelBase:
             raise NotImplementedError(f"Unable to convert filter: {expression=}")
         return regexp
 
+    def set_disable_fmu_logging_after_max_log_size(self, val: bool):
```
PeterMeisrimelModelon: Very open to better names for these functions.

filip-stenstrom: It could be an option and not a setter/getter?

tests/conftest.py
```diff
@@ -48,7 +48,7 @@ def download_url(url, save_file_to):
 
             with ZipFile(zip_path, 'r') as zf:
                 for fobj in zf.filelist:
-                    if fobj.filename.startswith('3.0') and fobj.filename.endswith('.fmu'):
+                    if (fobj.filename.startswith('1.0') or fobj.filename.startswith('3.0')) and fobj.filename.endswith('.fmu'):
```
filip-stenstrom, line 51: Why 1.0 as well?

tests/test_log.py
```diff
@@ -230,3 +251,47 @@ def test_resume_logging_on_increased_max_log_size(self):
...
+        fmu = FMUModelClass(fmu_path, log_level = 4)
+        assert fmu.get_disable_fmu_logging_after_max_log_size()
+        fmu.set_max_log_size(10)
+
+        assert not fmu.deactivate_called
+        fmu.append_log_message("Testing", 4, "additional logging")
+
+        assert fmu.has_reached_max_log_size()
+        assert fmu.deactivate_called
+        assert fmu.get_log_level() == 0
```
filip-stenstrom: Why do we need to check `deactivate_called`? Isn't it enough to verify that get_log_level() == 0, or look at produced log msg?

tests/test_log.py
```diff
@@ -17,16 +17,37 @@
...
 logs = os.path.join(file_path, "files", "Logs")
 
+this_dir = Path(__file__).parent.absolute()
+FMI1_REF_FMU_PATH = Path(this_dir) / 'files' / 'reference_fmus' / '1.0'
+FMI2_REF_FMU_PATH = Path(this_dir) / 'files' / 'reference_fmus' / '2.0'
+FMI3_REF_FMU_PATH = Path(this_dir) / 'files' / 'reference_fmus' / '3.0'
+
+def _get_fmu_class_logging_deactivation(parent_class):
+    class FMUModelLoggingDeactivation(parent_class):
+        """Testing class that injects some logging and test
```
filip-stenstrom: I don't see any logging injection?

CHANGELOG
```diff
@@ -9,6 +9,10 @@
       This allows setting reduced input/output update rates on a per-model basis.
     * Enabled use of `result_downsampling_factor` to `Master` algorithm on a per-model basis. 
     * Fixed an issue where `FMUModel(CS|ME)1.get_author()` returned a byte string, rather than a normal string.
+    * Logging will (by default) be explicitly be disabled by calling `fmiSetDebugLogging` 
+      and setting `log_level` to `0` when exceeding `set_max_log_size`.
+      * Use `fmu.set_disable_fmu_logging_after_max_log_size(<bool>)` to toggle this behavior.
+      * Use `fmu.get_disable_fmu_logging_after_max_log_size()` to query behavior, `default = True`.
```
filip-stenstrom: Why do we even need an option for this?

tests/test_log.py
```diff
@@ -230,3 +251,47 @@ def test_resume_logging_on_increased_max_log_size(self):
...
+
+    @pytest.mark.parametrize("parent_class, fmu_path",
+        [
+            (FMUModelME1, FMI1_REF_FMU_PATH / "me" / "Dahlquist.fmu"),
+            (FMUModelME2, FMI2_REF_FMU_PATH / "Dahlquist.fmu"),
+            (FMUModelME3, FMI3_REF_FMU_PATH / "Dahlquist.fmu"),
+        ]
+    )
+    def test_deactivate_logging_disable(self, parent_class, fmu_path):
+        """Test that logging is automatically disable upon reaching maximal log size."""
```
filip-stenstrom: Description seems wrong (copy-paste)?

src/pyfmi/fmi_base.pyx
```diff
@@ -397,6 +398,9 @@ cdef class ModelBase:
     def _get_module_name(self):
         return "Model"
 
+    def _deactivate_logging(self):
+        pass # to be implemented in class inheriting from ModelBase
```
efredriksson-modelon: So this should be an abstract method? Or at least raise a NotImplementedError.

tests/test_log.py
efredriksson-modelon: Agreed, having the tests decoupled from internal methods/implementation details would be nice here.

tests/test_log.py
```diff
@@ -17,16 +17,37 @@
...
+from pyfmi.fmi1 import FMUModelME1
+from pyfmi.fmi2 import FMUModelME2
+from pyfmi.fmi3 import FMUModelME3
 
 import numpy as np
 file_path = os.path.dirname(os.path.abspath(__file__))
 logs = os.path.join(file_path, "files", "Logs")
 
+this_dir = Path(__file__).parent.absolute()
+FMI1_REF_FMU_PATH = Path(this_dir) / 'files' / 'reference_fmus' / '1.0'
```
efredriksson-modelon: Minor, but calling `.absolute()` or  creating a new `Path` instance of `this_dir` is not needed.

CHANGELOG
efredriksson-modelon: I also wonder this. Trying to keep down the size of options/API would be good to reduce complexity. If we are really worried someone might need this at some point in the future or if we want it for testing then we can add an environment variable to toggle it.

tests/test_log.py
```diff
@@ -230,3 +251,47 @@ def test_resume_logging_on_increased_max_log_size(self):
...
         assert not fmu.has_reached_max_log_size()
         assert trunc_log_len < len(fmu.get_log()), "Logging did not appear to resume"
+
+    @pytest.mark.parametrize("parent_class, fmu_path",
+        [
+            (FMUModelME1, FMI1_REF_FMU_PATH / "me" / "Dahlquist.fmu"),
+            (FMUModelME2, FMI2_REF_FMU_PATH / "Dahlquist.fmu"),
+            (FMUModelME3, FMI3_REF_FMU_PATH / "Dahlquist.fmu"),
+        ]
+    )
```
efredriksson-modelon: Not strictly needed here, but it can be nice to parametrize a fixture instead of test: https://docs.pytest.org/en/stable/how-to/fixtures.html#fixture-parametrize, which can enable collecting common code in the fixture (like instantiating the FMU class). Feel free to skip.

CHANGELOG
PeterMeisrimelModelon: Removed it; can add later if needed.

src/pyfmi/fmi_base.pyx
PeterMeisrimelModelon: Switched to `NotImplementedError`

src/pyfmi/fmi_base.pyx
PeterMeisrimelModelon: Removed it; could have been an option, but would have made sense to have it on the FMU level, since the other functions relating to log size also exist on the FMUs.

tests/conftest.py
PeterMeisrimelModelon: Starting adding a test also for FMI1 with this PR. If this is "why are we starting to add tests for FMI1 now?", mostly since the added functionality also applies to FMI1 and it is straight-forward to add a test for all FMI versions.

tests/test_log.py
PeterMeisrimelModelon: Added, also in a few other places.

tests/test_log.py
PeterMeisrimelModelon: Removed the use of this class either way.

tests/test_log.py
PeterMeisrimelModelon: Skipping for now; we would need a separate pass over PyFMI testing as a whole, parametrizations and all. I propose we take this discussion when that day comes.

tests/test_log.py
PeterMeisrimelModelon: Fixed.

tests/test_log.py
PeterMeisrimelModelon: Removed the test, since we are skipping this functionality for now.

tests/conftest.py
filip-stenstrom: OK, but why not FMI 2.0?

PeterMeisrimelModelon: Ah, we also use FMI2 reference FMUs, but an earlier version, since the ones for 0.0.39 are currently broken (which has been reported as issue to reference FMUs repo).

That code is a bit lower down in the same file.
Hoping this can be uniform with reference FMUs 0.0.40.

filip-stenstrom (approved):

PR #366, opened 2025-10-29
Developer setup improvements

efredriksson-modelon:
This MR is to make it easier to develop and test PyFMI locally. It moves the CI/CD build setup over to a Docker image that you can run locally and a basic devcontainer setup based on it.

NOTE: As part of this I have removed testing for Python 3.9 in the CI/CD as this version is end of life.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,7 +4,7 @@ setup_requires =
     numpy >= 1.19.5
     cython >= 3.0
 
-python_requires = >=3.9
+python_requires = >=3.11
 
 install_requires =
     numpy >= 1.19.5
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -4,7 +4,7 @@ setup_requires =
     numpy >= 1.19.5
     cython >= 3.0
 
-python_requires = >=3.9
+python_requires = >=3.11
```
jschueller, line 7: why require >=3.11 ? 3.10 is not eol yet

efredriksson-modelon: @jschueller We run with 3.11 at Modelon so we don't need 3.10. It also enables using 3.11 features in the code.

Do you need 3.10? We could consider adding it but preferably only if there is a need for it.

jschueller: 3.10 is still supported on conda-forge, this will be breaking packages from that channel

efredriksson-modelon: I don't know about conda-forge but packaging for it hopefully have dependency handling on what versions PyFMI support and stopping it from getting released there for Python 3.10.

Do you know where this packing is done and if it is some person that might need to be pinged about this change?

PR #368, opened 2025-10-29
Getters for various FMI3 variable attributes

PeterMeisrimelModelon:
Largely the same as what we have for FMI2. 
Split `get_variable_nominal` into 2: by name and by valueref to be more pythonic.

Note: The use of Python enums here is subject to later re-evaluation.

### DIFF
--- a/src/pyfmi/fmi3.pyx
+++ b/src/pyfmi/fmi3.pyx
@@ -2641,7 +2641,7 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
         cdef FMIL3.fmi3_variability_enu_t variability = FMIL3.fmi3_import_get_variable_variability(variable)
         return FMI3_Variability(variability)
 
-    def get_variable_initial(self, variable_name) -> FMI3_Initial:
+    def get_variable_initial(self, variable_name: str) -> FMI3_Initial:
         """
         Get initial of the variable.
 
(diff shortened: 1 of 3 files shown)


tests/test_fmi3.py
```diff
@@ -76,6 +76,17 @@ def temp_dir_context(tmpdir):
...
+    def setup_class(cls):
+        cls._fmu_cache = {}
+
+    def _get_reference_fmu(self, name):
+        # Cache for FMU loading
+        if name not in self._fmu_cache:
+            self._fmu_cache[name] = load_fmu(FMI3_REF_FMU_PATH / (name + ".fmu"))
+        fmu = self._fmu_cache[name]
+        fmu.reset()
+        return fmu
```
efredriksson-modelon: We should avoid writing our own caches. This is also duplicated in 3 places, I would suggest breaking this out into a helper function that can be used in all places. Something like

```
@functools.cache
def _get_fmi3_reference_fmu_cached(name):
   ...

def get_fmi3_reference_fmu(name):
    fmu = _get_fmi3_reference_fmu_cached(name)
    fmu.reset()
    return fmu
```

src/pyfmi/fmi3.pyx
```diff
@@ -2597,6 +2603,270 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+
+            valueref --
+                The value reference of the variable.
+
+        Returns::
+
+            The name of the variable.
+
+        """
+        # Could have a better name?
```
efredriksson-modelon: I guess this comment is referencing `get_variable_by_valueref`? I would suggest `get_variable_name_by_valueref` so it is the same naming pattern as `get_variable_nominal_by_valueref` further down.

src/pyfmi/fmi3.pyx
```diff
@@ -2597,6 +2603,270 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+            value = FMIL3.fmi3_import_get_float32_variable_nominal(FMIL3.fmi3_import_get_variable_as_float32(variable))
+        else:
+            raise FMUException("Given variable type does not have a nominal.")
+        
+        if _override_erroneous_nominal and value <= 0:
+            variable_name = FMIL3.fmi3_import_get_variable_name(variable)
+            variablename = pyfmi_util.decode(variable_name)
+            if value == 0.0:
+                if self.callbacks.log_level >= FMIL.jm_log_level_warning:
+                    logging.warning("The nominal value for %s is 0.0 which is illegal according to the FMI specification. Setting the nominal to 1.0."%variablename)
```
efredriksson-modelon, line 2807: We should not log this on the system root logger. I would have expected this to go into the FMI logger?

tests/test_fmi3.py
```diff
@@ -828,12 +804,214 @@ def test_directional_derivatives(self):
...
+        err_msg = "The variable aaa could not be found."
+        with pytest.raises(FMUException, match = re.escape(err_msg)):
+            fmu.get_variable_initial("aaa")
+
+    def test_get_variable_min(self):
+        """Test get_variable_min."""
+        fmu = FMUModelME3(str(this_dir / "files" / "FMUs" / "XML" / "ME3.0" / "variableAttributes"),
+                          allow_unzipped_fmu = True, _connect_dll = False)
+        assert fmu.get_variable_min("float64") == -6.4
+        assert fmu.get_variable_min("float32") == np.float32(-3.2)
```
efredriksson-modelon: I guess this is a bigger question that is out of scope so asking mainly to learn/understand. Have we considered returning numpy objects instead of Python float? Now it seems like you get a Python float of 64-bit but the precision of it is actually worse.

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: I like this, added. Keeping the old variant around as an alias for backward-compatibility though.

tests/test_fmi3.py
PeterMeisrimelModelon: Good points, missed that here, since these are new with FMI3. Switched to using numpy.float32/64 resp. numpy.(u)intX these work somewhat seamless for any follow-up computations.

tests/test_fmi3.py
PeterMeisrimelModelon: Broke out into separate function(s), not moving to a separate place yet though. I think this is a good start though.

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: This is just copy-paste from FMI2 for now, I propose to revisit this type of logging on a separate occasion.

PR #369, opened 2025-10-31
fix: fixing some logging issues; checkpointing issue when disabling…

PeterMeisrimelModelon:
… logging; extra flush before xml_extraction

### DIFF
--- a/src/pyfmi/fmi_base.pyx
+++ b/src/pyfmi/fmi_base.pyx
@@ -212,13 +212,9 @@ cdef class ModelBase:
             alg = algorithm(start_time, final_time, input, self, options)
             # simulate
             alg.solve()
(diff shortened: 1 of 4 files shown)


src/pyfmi/fmi_base.pyx
```diff
@@ -438,6 +430,9 @@ cdef class ModelBase:
                 self._log_stream.seek(0)
             except AttributeError:
                 raise FMUException("In order to extract the XML-log from a stream, it needs to support 'seek'")
+        else:
+            if self.file_object:
```
PeterMeisrimelModelon, line 434: This ensures the `.txt` log is fully written out before trying to convert to `.xml`. Without this, there can be race conditions.

PR #370, opened 2025-11-04
Adding various unit getters for FMI3

PeterMeisrimelModelon:


src/pyfmi/fmi2.pyx
```diff
@@ -494,11 +494,7 @@ cdef class FMUModelBase2(FMI_BASE.ModelBase):
             A model as an object from the class FMUModelFMU2
         """
 
-        cdef int  status
```
PeterMeisrimelModelon, line 497: unused.

src/pyfmi/fmi2.pyx
```diff
@@ -2332,8 +2328,6 @@ cdef class FMUModelBase2(FMI_BASE.ModelBase):
         """
         cdef FMIL2.fmi2_import_variable_t* variable
         cdef FMIL2.fmi2_import_display_unit_t* display_unit
-        cdef FMIL2.fmi2_import_variable_typedef_t* variable_typedef
```
PeterMeisrimelModelon, line 2335: unused.

src/pyfmi/fmi3.pyx
```diff
@@ -2871,6 +2871,137 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+            String representing the display unit of the variable.
+        """
+        cdef FMIL3.fmi3_import_variable_t* variable
+        cdef FMIL3.fmi3_base_type_enu_t base_type
+        cdef bytes variable_name_bytes = pyfmi_util.encode(variable_name)
+        cdef char* variablename = variable_name_bytes
+        cdef FMIL3.fmi3_import_display_unit_t* display_unit
+        cdef const char* display_unit_name
+
+        display_unit = FMIL3.fmi3_import_get_variable_display_unit_by_name(self._fmu, variablename)
```
PeterMeisrimelModelon, line 2951: Note: Aliases can have distinct display units, which is why we added a convenience function in FMIL for this. Otherwise, this would require fetching the base variable and then iterating over the aliases,

src/pyfmi/fmi3.pyx
```diff
@@ -2871,6 +2871,137 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+        elif base_type == FMIL3.fmi3_base_type_float32:
+            unit = FMIL3.fmi3_import_get_float32_variable_unit(FMIL3.fmi3_import_get_variable_as_float32(variable))
+        else:
+            raise FMUException("Given variable type does not have units.")
+
+        if unit == NULL:
+            raise FMUException("No unit was found for the variable %s."%variable_name)
+
+        unit_name = FMIL3.fmi3_import_get_unit_name(unit)
+        return pyfmi_util.decode(unit_name) if unit_name != NULL else ""
```
MikaelJohnssonModelon, line 2928: ok

src/pyfmi/fmi3.pyx
MikaelJohnssonModelon: ok but they cannot have different values for the other attributtes (unit, quantity) so we only have to do this here?
So we have extra logic for alias in fmi3_import_get_variable_display_unit_by_name?

src/pyfmi/fmi3.pyx
```diff
@@ -3551,10 +3682,10 @@ cdef class FMUModelME3(FMUModelBase3):
         self._log_handler.capi_start_callback(self._max_log_size_msg_sent, self._current_log_size)
         status = FMIL3.fmi3_import_enter_initialization_mode(
             self._fmu,
-            tolerance_defined,
+            tol_defined,
             tolerance,
             start_time,
-            stop_time_defined,
+            stop_defined,
```
MikaelJohnssonModelon, line 3688: Bug fix?

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: Correct, an alias can have a different name, description and displayUnit (float types only), rest is identical.

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: Yes, we compute the correct values/types, but didn't actually use them.

PR #372, opened 2025-11-12
Enum support in fmi3

hugo-sigurdson:
Add support in FMI3 for enumerations.

### DIFF
--- a/src/pyfmi/fmi3.pxd
+++ b/src/pyfmi/fmi3.pxd
@@ -44,6 +44,14 @@ cdef class FMI3EventInfo:
     cdef public FMIL3.fmi3_float64_t nextEventTime
     # This will be populated further once we add support for CS and Clocks in particular.
 
+cdef class DeclaredType3:
+    cdef str _name
+    cdef str _description
+    cdef str _quantity
+
+cdef class EnumerationType3(DeclaredType3):
+    cdef dict _items
+
 cdef class FMUState3:
     """Class representing the FMU state, used with get and set FMU state. """
     cdef FMIL3.fmi3_FMU_state_t fmu_state
--- a/src/pyfmi/fmi3.pyx
+++ b/src/pyfmi/fmi3.pyx
@@ -170,6 +170,72 @@ cdef class FMI3EventInfo:
         self.nextEventTimeDefined              = FMIL3.fmi3_false
         self.nextEventTime                     = 0.0
 
+cdef class DeclaredType3:
+    """
+    Class defining data structure based on the XML element Type.
+    """
+    def __init__(self, name, description = "", quantity = ""):
+        self._name        = name
+        self._description = description
+        self._quantity = quantity
+
+    def _get_name(self):
+        """
+        Get the value of the name attribute.
+
+        Returns::
+
+            The name attribute value as string.
+        """
+        return self._name
+    name = property(_get_name)
+
+    def _get_description(self):
+        """
+        Get the value of the description attribute.
+
+        Returns::
+
+            The description attribute value as string (empty string if
+            not set).
+        """
+        return self._description
+    description = property(_get_description)
+
+cdef class EnumerationType3(DeclaredType3):
+    """
+    Class defining data structure based on the XML element Enumeration.
+    """
+    def __init__(self, name, description = "", quantity = "", items = None):
+        DeclaredType3.__init__(self, name, description, quantity)
+
+        self._items    = items
+
+    def _get_quantity(self):
+        """
+        Get the quantity of the enumeration type.
+
+        Returns::
+
+            The quantity as string (empty string if
+            not set).
+        """
+        return self._quantity
+    quantity = property(_get_quantity)
+
+    def _get_items(self):
+        """
+        Get the items of the enumeration type.
+
+        Returns::
+
+            The items of the enumeration type as a dict. The key is the
+            enumeration value and the dict value is a tuple containing
+            the name and description of the enumeration item.
+        """
+        return self._items
+    items = property(_get_items)
+
 cdef inline void _check_input_sizes(np.ndarray input_valueref, np.ndarray set_value):
     if np.size(input_valueref) != np.size(set_value):
         raise FMUException('The length of valueref and values are inconsistent. Note: Array variables are not yet supported')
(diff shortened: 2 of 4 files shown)


tests/test_fmi3.py
```diff
@@ -1459,6 +1459,23 @@ def test_get_description_of_alias(self):
...
 
+class Test_FMUModelBase3:
+    def test_declared_enumeration_type(self):
+        fmu = _get_fmu(FMI3_REF_FMU_PATH / "Feedthrough.fmu", _connect_dll = False)
+        enum = fmu.get_variable_declared_type("Enumeration_input")
+
+        assert len(enum.items.keys()) == 2
+        assert enum.items[1][0] == "Option 1"
+        assert enum.name == "Option"
+        assert enum.description == ""
```
efredriksson-modelon, line 1471: This looks a bit strange. We are testing `assert enum.name == "Option"` 2 times. Should we test
```
        assert enum.items[1][0] == "Option 1"
        assert enum.items[1][1] == "Option description"
```
?

It would probably be good to move the `enum.name` and `enum.description` assert up and here declare a varaible `item_1 = enum.items[1]` or similar that asserts are done on.

efredriksson-modelon (changes requested):
Nice work :+1: 

Just one thing around the tests that looks a bit strange.

tests/test_fmi3.py
hugo-sigurdson: Thanks; I split out the `enum` testing at the top and the item(s) at the bottom.  
Should hopefully make it more clear.

src/pyfmi/fmi3.pxd
```diff
@@ -44,6 +44,14 @@ cdef class FMI3EventInfo:
     cdef public FMIL3.fmi3_float64_t nextEventTime
     # This will be populated further once we add support for CS and Clocks in particular.
 
+cdef class DeclaredType3:
+    cdef object _name
```
PeterMeisrimelModelon: I think these 3 can actually the strings now,

src/pyfmi/fmi3.pyx
```diff
@@ -34,6 +34,7 @@ cimport pyfmi.fmil3_import as FMIL3
 cimport pyfmi.fmi_base as FMI_BASE
 cimport pyfmi.util as pyfmi_util
 from pyfmi.util import enable_caching
+from collections import OrderedDict
```
PeterMeisrimelModelon: Not sure if there is a point to specifically use `OrderedDict`. We have this in a lot of places in fmi2 code, because it was written before standard dicts guranteed to preserve the order.

src/pyfmi/fmi3.pxd
```diff
@@ -44,6 +44,14 @@ cdef class FMI3EventInfo:
...
+    cdef object _description
+    cdef object _quantity
+
+cdef class EnumerationType3(DeclaredType3):
+    cdef object _items
```
PeterMeisrimelModelon: could make this a dict as well.

src/pyfmi/fmi3.pyx
```diff
@@ -465,6 +532,78 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+        Returns::
+
+            The declared type.
+        """
+        cdef FMIL3.fmi3_import_variable_t* variable
+        cdef FMIL3.fmi3_value_reference_t  vr
+        cdef FMIL3.fmi3_import_variable_typedef_t* variable_type
+        cdef FMIL3.fmi3_base_type_enu_t    basetype
+        cdef FMIL3.fmi3_import_enumeration_typedef_t * enumeration_type
+        cdef FMIL3.fmi3_import_integer_typedef_t * integer_type
```
PeterMeisrimelModelon: A couple of these are unused for now, I propose to have a pass over these and remove any not (yet) used.

src/pyfmi/fmil3_import.pxd
```diff
@@ -237,6 +237,22 @@ cdef extern from 'fmilib.h':
         pass
     ctypedef fmi3_xml_alias_variable_t fmi3_import_alias_variable_t
 
+    cdef struct fmi3_xml_type_definitions_t:
```
PeterMeisrimelModelon: A few unused until further development, I propose to the remove those unused for now. Also, some names changes here for FMI3, so we can revisit using the correct ones when we actually start using them.

src/pyfmi/fmil3_import.pxd
```diff
@@ -400,6 +416,7 @@ cdef extern from 'fmilib.h':
     fmi3_import_enum_variable_t*    fmi3_import_get_variable_as_enum   (fmi3_import_variable_t*)
     fmi3_import_string_variable_t*  fmi3_import_get_variable_as_string (fmi3_import_variable_t*)
     fmi3_import_bool_variable_t*    fmi3_import_get_variable_as_boolean(fmi3_import_variable_t*)
+    fmi3_import_string_variable_t * fmi3_import_get_variable_as_string (fmi3_import_variable_t*)
```
PeterMeisrimelModelon: Duplicated? Already have that one 2 lines above.

Also, same for a few further down, maybe just check that no duplicated ones are added with this PR.

(The ordering of these compared to FMI2 changed, since it was largely unordered in FMI2)

tests/test_fmi3.py
```diff
@@ -1459,6 +1459,26 @@ def test_get_description_of_alias(self):
         assert self.fmu.get_variable_description("v5_a1") == ""
         assert self.fmu.get_variable_description("v5_a2") == "v5_a2_desc"
 
+class Test_FMUModelBase3:
+    def test_declared_enumeration_type(self):
```
PeterMeisrimelModelon, line 1463: Test looks good, but I'd prefer to have it split into 2:
1. Testing the valid enum variable.
2. Testing the non-existing variable + also checking the actual exception message.

+ an additional test for a variable that exists but does not have a declared type, e.g., an ordinary float64 variable.

See e.g., `test_get_variable_unit`, `test_get_variable_unit_when_no_unit`, `test_get_variable_unit_invalid_basetype`.

src/pyfmi/fmi3.pyx
```diff
@@ -465,6 +531,70 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+        cdef FMIL3.fmi3_string_t type_quantity
+        cdef FMIL3.fmi3_string_t item_desc
+        cdef FMIL3.fmi3_string_t item_name
+
+        variable_name = pyfmi_util.encode(variable_name)
+        cdef char* variablename = variable_name
+
+        variable = FMIL3.fmi3_import_get_variable_by_name(self._fmu, variablename)
+        if variable == NULL:
+            raise FMUException("The variable %s could not be found."%variablename)
```
PeterMeisrimelModelon: By having `variable_name = pyfmi_util.encode(variable_name)`, `variable_name` becomes a `byte` string, which makes the string formatting a bit messy here. 

I recommend to instead use 

```
variable_name_bytes = pyfmi_util.encode(variable_name)
cdef char* variablename = variable_name_bytes
```
and do the string formatting of both Exception messages using `variable_name` instead.

src/pyfmi/fmil3_import.pxd
```diff
@@ -433,12 +437,20 @@ cdef extern from 'fmilib.h':
     fmi3_initial_enu_t fmi3_import_get_variable_initial(fmi3_import_variable_t*)
     fmi3_string_t fmi3_import_get_variable_description(fmi3_import_variable_t*)
     int fmi3_import_get_variable_has_start(fmi3_import_variable_t*)
+    int fmi3_import_get_enum_type_item_value(fmi3_import_enumeration_typedef_t*, unsigned int)
     fmi3_import_variable_t* fmi3_import_get_variable(fmi3_import_variable_list_t* vl, size_t index)
     fmi3_import_variable_t* fmi3_import_get_variable_by_name(fmi3_import_t*, char*)
     fmi3_import_variable_t* fmi3_import_get_variable_by_vr(fmi3_import_t*, fmi3_value_reference_t)
     fmi3_value_reference_t fmi3_import_get_variable_vr(fmi3_import_variable_t*)
     fmi3_base_type_enu_t fmi3_import_get_variable_base_type(fmi3_import_variable_t*)
-    char* fmi3_import_get_model_version(fmi3_import_t*)
```
PeterMeisrimelModelon, line 441: Mistake I suppose? Should keep that one.

tests/test_fmi3.py
```diff
@@ -1459,6 +1459,37 @@ def test_get_description_of_alias(self):
...
+
+        assert item_1[0] == "Option 1"
+        assert item_1[1] == "First option"
+
+        assert item_2[0] == "Option 2"
+        assert item_2[1] == "Second option"
+
+    def test_non_existing_enumeration_type(self):
+        fmu = _get_fmu(FMI3_REF_FMU_PATH / "Feedthrough.fmu", _connect_dll = False)
+        err = "The variable b'non_existing_enumeration_input' could not be found."
```
PeterMeisrimelModelon: With the proposed change from the other comment this would simply be 

`err = "The variable non_existing_enumeration_input could not be found."`

tests/test_fmi3.py
```diff
@@ -1459,6 +1459,37 @@ def test_get_description_of_alias(self):
...
+
+        with pytest.raises(FMUException, match = re.escape(err)):
+            fmu.get_variable_declared_type("non_existing_enumeration_input")
+
+    def test_declared_variable_with_no_type(self):
+        fmu = _get_fmu(FMI3_REF_FMU_PATH / "Feedthrough.fmu", _connect_dll = False)
+        err = "The variable b'String_input' does not have a declared type."
```
PeterMeisrimelModelon: Same here.

PeterMeisrimelModelon (approved):

src/pyfmi/fmil3_import.pxd
efredriksson-modelon: We are missing a test to get the version of an FMI 3 FMU then?

src/pyfmi/fmi3.pyx
efredriksson-modelon: Standard dicts does not guarantee order, this is an implementation detail of cPython which could change. It is unlikely but we should use OrderedDict if order of the dict is important.

efredriksson-modelon: https://docs.python.org/3.7/whatsnew/3.7.html#:~:text=the%20insertion%2Dorder%20preservation%20natu%E2%80%A6, seems like this is old wisdom and it is now in fact part of the Python language spec.

PR #373, opened 2025-11-14
Set enum values via name

hugo-sigurdson:
Inspiration taken from `FMUModelBase2::_set -> enum`  

Hopefully the tests are self-explanatory enough.

### DIFF
--- a/tests/test_fmi3.py
+++ b/tests/test_fmi3.py
@@ -1490,6 +1490,46 @@ class Test_FMUModelBase3:
         with pytest.raises(FMUException, match = re.escape(err)):
             fmu.get_variable_declared_type("String_input")
 
+    def test_set_enum_by_string_name(self):
+        fmu = _get_fmu(FMI3_REF_FMU_PATH / "Feedthrough.fmu", _connect_dll = True)
+        assert fmu.get("Enumeration_input")[0] != np.int64(2)
+        fmu.set_enum(["Enumeration_input"], ["Option 2"])
+        assert fmu.get("Enumeration_input") == [2]
+
+    def test_set_enum_invalid_string(self):
+        fmu = _get_fmu(FMI3_REF_FMU_PATH / "Feedthrough.fmu", _connect_dll = False)
+        with pytest.raises(FMUException) as exc_info:
+            fmu.set_enum(["Enumeration_input"], ["invalid val"])
+        assert "not in the list of allowed enumeration items" in str(exc_info.value)
+        assert "Allowed values: Option 1, Option 2" in str(exc_info.value)
+
+    def test_set_enum_by_value_reference_integer(self):
+        fmu = _get_fmu(FMI3_REF_FMU_PATH / "Feedthrough.fmu", _connect_dll=True)
+        vref = fmu.get_variable_valueref("Enumeration_input")
+        fmu.set_enum([vref], [2])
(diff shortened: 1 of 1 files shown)


src/pyfmi/fmi3.pyx
```diff
@@ -1144,25 +1144,73 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+        variable_names = []
+        for value in valueref:
+            if isinstance(value, str) or isinstance(value, bytes):
+                # It's a variable name, get the value reference
+                value_int = self.get_variable_valueref(value)
+                processed_valuerefs.append(value_int)
+                variable_names.append(value)
+            else:
+                processed_valuerefs.append(int(value))
+                variable_names.append(None)
```
PeterMeisrimelModelon: Can use  `get_variable_name_by_valueref` here

src/pyfmi/fmi3.pyx
```diff
@@ -1144,25 +1144,73 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
+
+        if len(valueref) != len(values):
+            raise FMUException("Enum references and values must have the same length")
+
+        processed_values = []
+        for i, value in enumerate(values):
+            if isinstance(value, (str, bytes)):
+                if variable_names[i] is None:
```
PeterMeisrimelModelon: With above comment, we can safely assume it won't be `None`

tests/test_fmi3.py
```diff
@@ -1490,6 +1490,57 @@ def test_declared_variable_with_no_type(self):
         with pytest.raises(FMUException, match = re.escape(err)):
             fmu.get_variable_declared_type("String_input")
 
+    def test_set_enum_by_string_name(self):
+        fmu = _get_fmu(FMI3_REF_FMU_PATH / "Feedthrough.fmu", _connect_dll = True)
+        fmu.set_enum(["Enumeration_input"], ["Option 1"])
```
PeterMeisrimelModelon: `Option 1` might just be the default value? Should add an `assert fmu.get("Enumeration_input")[0] != 1` before that to ensure that `set` actually works.

Same for some more tests below.

tests/test_fmi3.py
```diff
@@ -1490,6 +1490,57 @@ def test_declared_variable_with_no_type(self):
...
+        assert fmu.get("Enumeration_input") == [1]
+
+    def test_set_enum_invalid_string(self):
+        fmu = _get_fmu(FMI3_REF_FMU_PATH / "Feedthrough.fmu", _connect_dll = True)
+        with pytest.raises(FMUException) as exc_info:
+            fmu.set_enum(["Enumeration_input"], ["invalid val"])
+        assert "not in the list of allowed enumeration items" in str(exc_info.value)
+        assert "Allowed values:" in str(exc_info.value)
```
PeterMeisrimelModelon: Can we test that the valid options are correctly displayed? 
With the encoding/decoding it is easy to make mistakes here.

PeterMeisrimelModelon (changes requested):
Looks good! A few minor comments to further improve.

tests/test_fmi3.py
hugo-sigurdson: Good catch, changing the test to another enum value.

PR #374, opened 2025-11-20
New ResultHandler with diagnostics data in data_3

hugo-sigurdson:
New `ResultHandlerCalcDiags(ResultReader)` with diagnostics data written explicitly into data_3.  
This also comes with a few test cases and a QoL update to the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ build-dev-image:
 build: .venv
 	$(call _run, python setup.py install --fmil-home=/usr)
 
-test: .venv
+test: build
 	$(call _run, pytest)
 
 shell:


src/common/io.py
```diff
@@ -2990,6 +2990,163 @@ def get_result(self) -> ResultDymolaBinary:
...
+            variable_index = self._get_variable_name_to_index_dict().get(variable_name)
+            data_mat = self._dataInfo[0][variable_index]
+            data_index = abs(self._dataInfo[1][variable_index]) - 1 # abs due to alias
+
+            if data_mat == 0:
+                data_mat = 2 if len(self.raw['data_2']) > 0 else 1
+
+            return data_index, data_mat
+
+        raise KeyError("Could not find variable")
```
efredriksson-modelon: Should this not be a VariableNotFoundError exception?

Also, should have a test for hitting this exception.

src/common/io.py
```diff
@@ -2990,6 +2990,163 @@ def get_result(self) -> ResultDymolaBinary:
...
+        name_dict: dict = fmi_util.read_name_list(
+            encode(self._fname),
+            self._name_info["file_position"],
+            int(self._name_info["nbr_variables"]),
+            int(self._name_info["max_length"])
+        )
+
+        return {decode(k): v for k, v in name_dict.items()}
+
+    @cache
```
efredriksson-modelon: I think we can skip cache for this one as it is just calling another function that is cached. Having all the names 2 times in memory is probably not good as time saved will be negligible.

src/common/io.py
```diff
@@ -2990,6 +2990,163 @@ def get_result(self) -> ResultDymolaBinary:
...
+                    if self._contains_diagnostic_data
+                    else self._get_trajectory(data_index)
+                )
+                return Trajectory(time, data)
+            case 3:
+                return Trajectory(time, self._get_diagnostics_trajectory(data_index))
+            case _:
+                raise ValueError(f"Invalid data matrix: {data_mat}")
+
+    @cache
```
efredriksson-modelon: Why does all the get_trajectory functions have caching? I would prefer if they did not have caching as to keep things more flexible, caching for trajectories should probably not be at this level. I guess we might want caching for the 'time' variable as it is used many times internally. Maybe we can create separate methods for getting the time variables and add cache on them?

src/common/io.py
```diff
@@ -2990,6 +2990,163 @@ def get_result(self) -> ResultDymolaBinary:
...
+                }
+            case _:
+                return super().read_sub_array(hdr, copy)
+
+class _DelayedVariableLoadDiags(DelayedVariableLoad):
+    def initialize_read(self):
+        self.dtypes = convert_dtypes(mdtypes_template, self.byte_order)
+        self._matrix_reader = _DelayedVarReader4Diags(self)
+
+class ResultHandlerCalcDiags(ResultReader):
```
efredriksson-modelon: This class should be private and only used as delegate in the class we add in the next step?

I would suggest a name like:
`ResultHandlerCalcDiags` -> `_ResultHandlerBinaryMatConsolidated` as it uses a simplified table layout created after a simulation.

And the tests can import it as private until we add the delegator in the next step.

tests/test_io.py
```diff
@@ -2355,3 +2357,79 @@ def test_basic_variable_get_set_result_correctness(fmu_path, result_handling):
...
+        result = ResultHandlerCalcDiags(mat_file)
+        assert result.get_trajectory("@Diagnostics.step_time") is not None
+
+        path = tmp_path / "mock_result_data_no_diag.mat"
+        variable_names = ["time", "spring.phi_nominal", "torque.flange.phi"]
+        data_info = np.zeros((2, len(variable_names)), dtype=np.int32)
+        _save_mat_file(path, data_info, {"data_2": np.zeros((4, 3))}, variable_names)
+
+        result_no_diag = ResultHandlerCalcDiags(str(path))
+        assert "@Diagnostics.step_time" not in result_no_diag.get_variable_names()
```
efredriksson-modelon: This could be split out into a fixture and a separate test?

Generally I think these tests would be easier to get an overview of if all result files created where fixtures.

tests/test_io.py
```diff
@@ -2355,3 +2357,79 @@ def test_basic_variable_get_set_result_correctness(fmu_path, result_handling):
...
+    variable_names = ["time", "spring.phi_nominal", "torque.flange.phi", "@Diagnostics.step_time"]
+    data_info = np.array([[3, 2, 2, 3], [1, 2, 3, 2]], dtype=np.int32)
+
+    time_data_2 = np.arange(3, dtype=float)
+    data_2 = np.vstack([time_data_2, [1, 2, 3], [1, 2, 3], [1, 2, 3]])
+
+    time_data_3 = np.arange(1, 4, dtype=float)
+    data_3 = np.vstack([time_data_3, [1, 2, 3], [1, 2, 3]])
+
+    return _save_mat_file(path, data_info, {"data_2": data_2, "data_3": data_3}, variable_names)
```
efredriksson-modelon: One idea to make things simpler could be to have `_save_mat_file` take `tmp_path` and in `_save_mat_file` generate a random path like tmp_path / (uuid.uuid4().hex + '.mat') so you never have to think about constructing paths manually for all fixtures.

tests/test_io.py
```diff
@@ -2355,3 +2357,79 @@ def test_basic_variable_get_set_result_correctness(fmu_path, result_handling):
...
+
+@pytest.fixture(name="mat_file")
+def mat_file_fixture(tmp_path):
+    path = tmp_path / "mock_result_data.mat"
+
+    variable_names = ["time", "spring.phi_nominal", "torque.flange.phi", "@Diagnostics.step_time"]
+    data_info = np.array([[3, 2, 2, 3], [1, 2, 3, 2]], dtype=np.int32)
+
+    time_data_2 = np.arange(3, dtype=float)
+    data_2 = np.vstack([time_data_2, [1, 2, 3], [1, 2, 3], [1, 2, 3]])
```
efredriksson-modelon: If we add more fixtures one thing to experiment with would be to have them be single expression:
```
return _save_mat_file(...,
     data_info=np.array([[3, 2, 2, 3], [1, 2, 3, 2]], dtype=np.int32),
     data_dict={
            'data_2': np.vstack([[0, 1, 2], [1, 2, 3], [1, 2, 3], [1, 2, 3]])
            ...
```

which could improve readability of the fixtures. But see this as a suggestion, will depend on what we put in for data.

tests/test_io.py
```diff
@@ -2355,3 +2357,79 @@ def test_basic_variable_get_set_result_correctness(fmu_path, result_handling):
...
+        data_info = np.array([[2, 2], [1, 2]], dtype=np.int32)
+
+        time_data = np.array([0], dtype=float)
+        data_2 = np.vstack([time_data, [42]])
+
+        _save_mat_file(path, data_info, {"data_2": data_2}, variable_names)
+
+        result = ResultHandlerCalcDiags(str(path))
+        traj = result.get_trajectory("spring.phi_nominal")
+        assert traj.x == 42
```
efredriksson-modelon: This looks a bit strange, it should not be a numpy array? Also, do we want a test where we assert trajectory values for other cases? No diagnostics, interpolation?

src/common/io.py
PeterMeisrimelModelon: +1 on this, with the addition of naming it `_ResultReader*` Handlers are the writers right now.

src/common/io.py
```diff
@@ -2990,6 +2990,163 @@ def get_result(self) -> ResultDymolaBinary:
...
+    def __init__(self, fname):
+        """
+        Load a .mat result file with diagnostics data explicitly written to data_3.
```
PeterMeisrimelModelon: This docstring only makes sense in the context of the exported format of `ResultHandlerBinaryFile`. I propose a reference or to make it understandable in isolation.

src/common/io.py
PeterMeisrimelModelon: Note: We have this type of caching in `ResultDymolaBinary`, but it does not make too much sense on reflection, except for time.

tests/test_io.py
```diff
@@ -2355,3 +2357,79 @@ def test_basic_variable_get_set_result_correctness(fmu_path, result_handling):
...
+    mat_dict = {"dataInfo": data_info, "name": _convert_char_array(variable_names), **data_dict}
+    savemat(path, mat_dict, format="4")
+    return path
+
+@pytest.fixture(name="mat_file")
+def mat_file_fixture(tmp_path):
+    path = tmp_path / "mock_result_data.mat"
+
+    variable_names = ["time", "spring.phi_nominal", "torque.flange.phi", "@Diagnostics.step_time"]
+    data_info = np.array([[3, 2, 2, 3], [1, 2, 3, 2]], dtype=np.int32)
```
PeterMeisrimelModelon: This one doesn't seems quite right, the base line for every result should be:

* time as index 1 in data matrix 2; the ordinary time for continuous variables
* @Diagnostics.step_time as index 1 in data matrix 3; diagnostics time

`spring.phi_nominal`, which I presume is to represent a constant/parameter (and can maybe be named as such), should be in data matrix 1.

Other than that I recommend to have at least 2 variables for each data matrix, with different values, so that we have some solid values that we can test against to see that the variable_name -> index mapping works.

tests/test_io.py
```diff
@@ -2355,3 +2357,79 @@ def test_basic_variable_get_set_result_correctness(fmu_path, result_handling):
...
+
+    time_data_3 = np.arange(1, 4, dtype=float)
+    data_3 = np.vstack([time_data_3, [1, 2, 3], [1, 2, 3]])
+
+    return _save_mat_file(path, data_info, {"data_2": data_2, "data_3": data_3}, variable_names)
+
+class TestResultHandlerForCalculatedDiagnostics:
+    def test_get_all_variable_names(self, mat_file):
+        result = ResultHandlerCalcDiags(mat_file)
+        assert len(result.get_variable_names()) == 4
```
PeterMeisrimelModelon: I recommend to check the expected names are there.

tests/test_io.py
```diff
@@ -2355,3 +2357,79 @@ def test_basic_variable_get_set_result_correctness(fmu_path, result_handling):
...
+        traj = result.get_trajectory("spring.phi_nominal")
+        assert traj.x == 42
+
+    def test_empty_mat_file(self, tmp_path):
+        path = tmp_path / "mock_result_data_empty.mat"
+        savemat(str(path), {}, format="4")
+
+        with pytest.raises(Exception) as exception:
+            result = ResultHandlerCalcDiags(str(path))
+            assert "buffer is too small for requested array" in str(exception.value)
```
PeterMeisrimelModelon: Where does this one come from?

tests/test_io.py
```diff
@@ -2355,3 +2357,79 @@ def test_basic_variable_get_set_result_correctness(fmu_path, result_handling):
...
+            assert result.get_variable_names() == []
+            assert result.get_trajectory("nonexistent") is None
```
PeterMeisrimelModelon: Shouldn't this be a `ValueError` or similar?

src/common/io.py
```diff
@@ -2990,6 +2990,178 @@ def get_result(self) -> ResultDymolaBinary:
...
+    @cached_property
+    def _time_vector(self) -> np.ndarray:
+        return fmi_util.read_trajectory(
+            encode(self._fname),
+            0,
+            self._data_2_info["file_position"],
+            self._data_2_info["sizeof_type"],
+            int(self._data_2_info["nbr_points"]),
+            int(self._data_2_info["nbr_variables"])
+        )
```
efredriksson-modelon: `self._get_trajectory(0)` can be used

src/common/io.py
```diff
@@ -2990,6 +2990,178 @@ def get_result(self) -> ResultDymolaBinary:
...
+    @cached_property
+    def _diagnostics_time_vector(self) -> np.ndarray:
+        return fmi_util.read_trajectory(
+            encode(self._fname),
+            0,
+            self._data_3_info["file_position"],
+            self._data_3_info["sizeof_type"],
+            int(self._data_3_info["nbr_points"]),
+            int(self._data_3_info["nbr_variables"])
+        )
```
efredriksson-modelon: Same, using _get_diagnostics_trajectory.

src/common/io.py
efredriksson-modelon: Right, should be Reader not Handler.

tests/test_io.py
efredriksson-modelon: Ah, I guess this works because `assert np.array([[42]]) == 42` seem to work for example. Not because `x` is a `float`.

tests/test_io.py
hugo-sigurdson: The error occurs because the test creates an empty .mat file with no data, so when scipy tries to read the file header it cannot find the required 4 bytes to determine the byte order (?). Looks like since the buffer returned from the empty file is smaller than expected, NumPy raises the error.

**Full stacktrace**
```python
/src/.venv/lib/python3.11/site-packages/pyfmi/common/io.py:3052: in __init__
    delayed = _DelayedVariableLoadDiags(f, chars_as_strings=False)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/usr/local/lib/python3.11/dist-packages/scipy/io/matlab/_mio4.py:325: in __init__
    super().__init__(mat_stream, *args, **kwargs)
/usr/local/lib/python3.11/dist-packages/scipy/io/matlab/_miobase.py:382: in __init__
    byte_order = self.guess_byte_order()
                 ^^^^^^^^^^^^^^^^^^^^^^^
/usr/local/lib/python3.11/dist-packages/scipy/io/matlab/_mio4.py:330: in guess_byte_order
    mopt = read_dtype(self.mat_stream, np.dtype('i4'))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

We might want to align this with what @efredriksson-modelon spoke on, in regards to an exception type for when a result does not exist.

tests/test_io.py
hugo-sigurdson: Not too certain that the two (maybe three) last asserts give us anything, considering if the file is empty then surely no variables or data can exist.

hugo-sigurdson: But it does throw a valid exception if the file contains data.

> pyfmi.common.io.VariableNotFoundError: Cannot find variable 'nonexistent' in data file.

tests/test_io.py
efredriksson-modelon: I think the test is confusing because there are multiple lines under `with pytest.raises`, we should probably aim at only having the single line that throws the exception to make things clear. I assume right now it is one of the asserts that are not checked because an exception is throw as part of that line?

For exception when the .mat file is corrupt we can probably wait for the next part when we do the delegation, I think as part of that we will have to check the fields in the .mat file and handle if it is corrupt.

src/common/io.py
```diff
@@ -2990,6 +2990,169 @@ def get_result(self) -> ResultDymolaBinary:
...
+            variable_index = self._get_variable_name_to_index_dict().get(variable_name)
+            data_mat = self._dataInfo[0][variable_index]
+            data_index = abs(self._dataInfo[1][variable_index]) - 1 # abs due to alias
+
+            if data_mat == 0:
+                data_mat = 2 if len(self.raw['data_2']) > 0 else 1
+
+            return data_index, data_mat
+
+        raise VariableNotFoundError(f"Cannot find variable '{variable_name}' in data file.")
```
efredriksson-modelon, line 3088: Test checking this exception throws?

tests/test_io.py
```diff
@@ -2355,3 +2358,103 @@ def test_basic_variable_get_set_result_correctness(fmu_path, result_handling):
...
+    phi_data = np.array([0.0, 10.0, 20.0], dtype=float)
+    data_2 = np.vstack([time_data, phi_data])
+    return _save_mat_file(tmp_path, data_info, {"data_2": data_2}, variable_names)
+
+class TestResultReaderForBinaryMatConsolidated:
+    def test_get_all_variable_names(self, mat_file):
+        result = _ResultReaderBinaryMatConsolidated(mat_file)
+        variables = result.get_variable_names()
+        assert len(variables) == 6
+        assert variables == ["spring.phi_nominal", "spring.k_constant", "time","torque.flange.phi", "@Diagnostics.step_time", "@Diagnostics.error_code"]
```
PeterMeisrimelModelon: Minor nitpick; I would use sets as to not regression check the exact ordering.

tests/test_io.py
```diff
@@ -2355,3 +2358,103 @@ def test_basic_variable_get_set_result_correctness(fmu_path, result_handling):
...
+        assert result.get_trajectory("@Diagnostics.step_time") is not None
+
+    def test_without_diagnostic_variable(self, mat_file_no_diag):
+        result_no_diag = _ResultReaderBinaryMatConsolidated(mat_file_no_diag)
+        assert "@Diagnostics.step_time" not in result_no_diag.get_variable_names()
+
+    def test_edge_case_single_solver_point(self, mat_file_single_point):
+        result = _ResultReaderBinaryMatConsolidated(mat_file_single_point)
+        traj = result.get_trajectory("spring.phi_nominal")
+        assert traj.x == np.array(42)
```
PeterMeisrimelModelon:
```suggestion
        assert traj.x == np.array([42])
```

numpy.array singletons, i.e., no lists can do funny things, would try to avoid.

tests/test_io.py
PeterMeisrimelModelon: It sounds like there should be an try: except: ... on the `_DelayedVariableLoadDiags` call and re-raise some sort of "invalid/corrupt mat format" exception?

tests/test_io.py
PeterMeisrimelModelon: Maybe we should not have multiple calls within a with `pytest.raises` context, then it is more clear what and what doesn't cause exceptions.

tests/test_io.py
```diff
@@ -2355,3 +2358,103 @@ def test_basic_variable_get_set_result_correctness(fmu_path, result_handling):
...
+    data_info = np.zeros((2, len(variable_names)), dtype=np.int32)
+    return _save_mat_file(tmp_path, data_info, {"data_2": np.zeros((4, 3))}, variable_names)
+
+@pytest.fixture
+def mat_file_single_point(tmp_path):
+    variable_names = ["time", "spring.phi_nominal"]
+    data_info = np.array([[2, 2], [1, 2]], dtype=np.int32)
+    time_data = np.array([0], dtype=float)
+    data_2 = np.vstack([time_data, [42]])
+    return _save_mat_file(tmp_path, data_info, {"data_2": data_2}, variable_names)
```
PeterMeisrimelModelon: The single point issue is primarily an edge case for interpolation, which we only do in the context of interpolating data_2 onto the data_3 time trajectory. So we do not data_3 here to test this.

tests/test_io.py
```diff
@@ -2355,3 +2358,103 @@ def test_basic_variable_get_set_result_correctness(fmu_path, result_handling):
...
+
+@pytest.fixture
+def mat_file_empty(tmp_path):
+    path = tmp_path / "mock_result_data_empty.mat"
+    savemat(str(path), {}, format="4")
+    return str(path)
+
+@pytest.fixture
+def mat_file_interpolation(tmp_path):
```
PeterMeisrimelModelon, line 2399: See comment above, interpolation is only relevant when there is data_3, so we need data_3 to test this. 

In particular, we also need data_3 to be of different length than ordinary data_2 time.

tests/test_io.py
```diff
@@ -2355,3 +2358,103 @@ def test_basic_variable_get_set_result_correctness(fmu_path, result_handling):
...
+
+    def test_empty_mat_file(self, mat_file_empty):
+        with pytest.raises(TypeError) as exception:
+            result = _ResultReaderBinaryMatConsolidated(mat_file_empty)
+            assert "buffer is too small for requested array" in str(exception.value)
+            assert result.get_variable_names() == []
+            assert result.get_trajectory("nonexistent") is None
+
+    def test_interpolation_between_points(self, mat_file_interpolation):
+        result = _ResultReaderBinaryMatConsolidated(mat_file_interpolation)
```
PeterMeisrimelModelon: This one is odd, there should not be any interpolation when we do not have data_3

src/common/io.py
hugo-sigurdson: This is covered by the test `test_get_all_non_existing_variable_throws`.

PR #402, opened 2026-04-08
chore: enable parallel c compilation for cython build

PeterMeisrimelModelon:
In CI, this approximately halves "build" time. Local re-building, this can be up to a factor of 4.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 .PHONY: build build-dev-image test shell
 DOCKER_IMAGE := pyfmi-dev
 IN_DOCKER_IMG := $(shell test -f /.dockerenv && echo 1 || echo 0)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,6 +1,7 @@
 .PHONY: build build-dev-image test shell
 DOCKER_IMAGE := pyfmi-dev
 IN_DOCKER_IMG := $(shell test -f /.dockerenv && echo 1 || echo 0)
+MAKE_JFLAG=-j$(shell nproc)
```
efredriksson-modelon: I think this option is related to setuptools, so we could consider calling SETUPTOOLS_JFLAG if we want to be more explicit.

efredriksson-modelon (approved):

Issue #403, opened 2026-04-29
ME FMU without continuous states crashes


cbu-hw:
We have a ME FMU that (accidentally) had no continuous states, and simulating this crashes Python. The FMU simulates correctly using fmpy and OMSimulator.

The model:
```modelica
model ElastoGapMWE
  Modelica.Mechanics.Translational.Components.ElastoGap elastoGap(s_rel(start = 0), c = 100, d = 0)  annotation(
    Placement(transformation(origin = {26, 14}, extent = {{-10, -10}, {10, 10}})));
  Modelica.Mechanics.Translational.Components.Fixed fixed annotation(
    Placement(transformation(origin = {68, -4}, extent = {{-10, -10}, {10, 10}})));
  Modelica.Mechanics.Translational.Sources.Position position(exact = true)  annotation(
    Placement(transformation(origin = {-20, 14}, extent = {{-10, -10}, {10, 10}})));
  Modelica.Blocks.Sources.Sine sine(f = 1, continuous = true)  annotation(
    Placement(transformation(origin = {-70, 14}, extent = {{-10, -10}, {10, 10}})));
equation
  connect(position.flange, elastoGap.flange_a) annotation(
    Line(points = {{-10, 14}, {16, 14}}, color = {0, 127, 0}));
  connect(elastoGap.flange_b, fixed.flange) annotation(
    Line(points = {{36, 14}, {68, 14}, {68, -4}}, color = {0, 127, 0}));
  connect(sine.y, position.s_ref) annotation(
    Line(points = {{-59, 14}, {-32, 14}}, color = {0, 0, 127}));
  annotation(
    uses(Modelica(version = "4.1.0")),
  experiment(StartTime = 0, StopTime = 1, Tolerance = 1e-06, Interval = 0.002));
end ElastoGapMWE;
```
Export this as an ME FMU.

Python reproducer script:
```python
import faulthandler
import pyfmi
import assimulo

print(f'{pyfmi.__version__=}') # 2.21.0
print(f'{assimulo.__version__=}') # 3.7.3
faulthandler.enable()  # Enable Python to print a traceback on crashing

model = pyfmi.load_fmu('ElastoGapMWE.fmu', kind='ME', log_level=6)
print(model.simulate_options())
print(f'{model.continuous_states=}\n')

model.simulate(final_time=1.0) # crashes, following print never occurs
print("Finished simulation")
```
Run the Python script. Note that the `continuous_states` array is empty.
Observe a crash:
```
[...]

PR #404, opened 2026-04-29
fix: improved robustness in event handling for models without states

PeterMeisrimelModelon:
Resolves the PyFMI related issue in https://github.com/modelon-community/PyFMI/issues/403

### DIFF
--- a/tests/test_fmi2.py
+++ b/tests/test_fmi2.py
@@ -805,6 +805,18 @@ class Test_FMUModelME2:
 
         ET.parse(model.extract_xml_log()) # Exception if not well-formed XML
 
+    def test_nostate_with_event_update_continous_states_changed_flags_true(self):
+        """Test that models without states robustly handle 'event_update' returning
+        'nominalsOfContinuousStatesChanged' or 'valuesOfContinuousStatesChanged' == True."""
+        class FMUModelME2Test(FMUModelME2):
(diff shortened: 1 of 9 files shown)


tests/test_fmi2.py
```diff
@@ -805,6 +805,16 @@ def _get_A(self, *args, **kwargs):
 
         ET.parse(model.extract_xml_log()) # Exception if not well-formed XML
 
+    def test_nostate_inconsistent_event_update_output(self):
```
efredriksson-modelon: This test is a bit hard to understand what it exactly does. I assume we just want to test that we do not crash for nominalsOfContinuousStatesChanged=True when there are no states? Even if this flag does not really make sense here? A give/when/then naming pattern here could make the test clearer.

efredriksson-modelon (approved):
Looks good, just one thing on making the test clearer.

tests/test_fmi2.py
PeterMeisrimelModelon: Updates the name & description. 

Yes, this is an odd case, since the flags convey no meaning if there are no states. So it is legit that an FMU doesn't touch them at all, rather than explicitly setting them to false.

The fix contains the more robust solution to ignore them if there are no states.

efredriksson-modelon (approved):

PR #412, opened 2026-05-27
fix: handle state events for ME2/ME3 on consecutive simulations

### LINKED ISSUE #159: The FMU cannot have correct results with time-changing input
Hello, I'm using dymola-generated FMU for MPC work. I used PyFMI package's function to load the FMU and operate it.

Now my model is based on Modelica Buildings Library 9.0.0 and I found the issues shown below.

**Describe the bug**

I exported my system model based on MBL 9.0.0 to a co-simulation FMU for model predictive control.
There is only **one integer input** for the system mode (1,2 or 3).

I divided the simulation period into two parts (t1-t2, t2-t3) with different signals.

When my input signal ('Mode') was changed from 1 to 2 or 2 to 3, the results showed the system didn't have the correct reflection I wanted.

**Pseudocode:**

model.reset()
model.initialize()
model.set('Mode', int(2))
model.simulate(start_time=t1, final_time=t2,
input=None, options=demo_options)
states = model.get_fmu_state()

model_options['initialize'] = False
model.set_fmu_state(states)
model.set('Mode', int(3))
model.simulate(start_time=t2, final_time=t3,
input=None, options=demo_options)

The input 'Mode' is directly connected with one variable called 'TES'.
During the first time range from t1 to t2, the 'TES' is equal to input 'Mode'.
However, during the second time range from t2 to t3, the 'TES' is still the old value even though the input 'Mode' has been changed from 1 to 2 or 2 to 3 (this issue is shown in Figure 1).

The system didn't work correctly during the second time range either, which is very weird.
I tried different IDEs, Spyder or Pycharm, and Dymola versions, 2021, 2021X, and 2023. This issue still exists.

I also tried one example model (Buildings.Fluid.Storage.Examples.Stratified) in MBL 9.0.0, but the issue also exists.

And another strange stuff is that sometimes the 'TES' could follow the 'Mode', but some variables will restart from their initial guess (this issue is shown in Figure 2).

**Expected behavior**

Issue 1:

Throughout the simulation period, the variables in the system model should be influenced by the inputs.
Now they are influenced just in the first time range.

Issue 2:

The variables in the second time range should start from their previous values in the last time range instead of the initial guess.

**Screenshots**

Figure 1. The input cannot influence the system model except for the first time range.

Snipaste_2022-12-09_20-48-15

Figure 2. The variables restart from their initial guess.

Snipaste_2022-12-09_20-48-27

**Version**

PyFMI: 2.9.7
Modelica Buildings Library: 9.0.0
Modelica 4.0.0. Dymola 2021, 2021X, 2023
Windows 10

**Additional context**

I'm wondering if there are any comments or suggestions for this issue.
This issue is really weird.
I would appreciate any hints.
Thanks for your kind help!

jschueller:
When continuing a simulation with initialize=False (e.g. after set_fmu_state()
+ input changes), the FMU must process potential state events before resuming integration. The FMI spec requires calling fmi2NewDiscreteStates / fmi3UpdateDiscreteStates after fmi2SetFMUstate since restored state + new inputs may trigger events.

Previously, time-event checks at start time only covered ME1 and used unreliable event_info flags. PR #109's unconditional event_update() was too aggressive because it ran on the initialize=True path too, where event_update was already called during init, violating the FMI state machine.

This fix:
- For initialize=False with ME2/ME3: enter_event_mode + event_update + enter_continuous_time_mode to process pending events (e.g. after set_fmu_state + input changes)
- For initialize=True with ME2/ME3: extend the existing time-event check (equivalent to the ME1 check that was already in place) with proper mode transition wrappers
- ME1 behavior unchanged
- Add enter_event_mode() no-op override to Dummy_FMUModelME2 to prevent segfault when algorithm driver calls enter_event_mode() on test stubs

Fixes #159

Tests added:
- test_fmi2.py::test_consecutive_simulation_with_initialize_false
- test_fmi3_sim.py::TestSimulation::test_consecutive_simulation_with_initialize_false

### DIFF
--- a/src/pyfmi/fmi_algorithm_drivers.py
+++ b/src/pyfmi/fmi_algorithm_drivers.py
@@ -374,11 +374,22 @@ class AssimuloFMIAlg(AlgorithmBase):
                                                                                  solver_options = self.solver_options)
             number_of_diagnostics_variables = len(_diagnostics_vars)
 
-        #See if there is an time event at start time
+        #Check for events at start time
         if isinstance(self.model, FMUModelME1):
             event_info = self.model.get_event_info()
             if event_info.upcomingTimeEvent and event_info.nextEventTime == model.time:
                 self.model.event_update()
+        elif isinstance(self.model, (FMUModelME2, CoupledFMUModelME2, FMUModelME3)):
+            if not self.options['initialize']:
+                self.model.enter_event_mode()
+                self.model.event_update()
+                self.model.enter_continuous_time_mode()
+            else:
+                event_info = self.model.get_event_info()
+                if event_info.nextEventTimeDefined and abs(event_info.nextEventTime - model.time) <= 1e-14:
+                    self.model.enter_event_mode()
+                    self.model.event_update()
+                    self.model.enter_continuous_time_mode()
 
         if abs(start_time - model.time) > 1e-14:
             logging_module.warning('The simulation start time (%f) and the current time in the model (%f) is different. Is the simulation start time correctly set?'%(start_time, model.time))
--- a/src/pyfmi/test_util.pyx
+++ b/src/pyfmi/test_util.pyx
@@ -396,6 +396,9 @@ class Dummy_FMUModelME2(_ForTestingFMUModelME2):
     def event_update(self, *args, **kwargs):
         pass
 
+    def enter_event_mode(self, *args, **kwargs):
+        pass
+
     def enter_continuous_time_mode(self, *args, **kwargs):
         pass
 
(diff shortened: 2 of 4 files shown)


jschueller:
hello @modelonrobinandersson @PeterMeisrimelModelon, what do you think ?

PeterMeisrimelModelon:
@jschueller Sorry for the late response, I am aware of the various MRs, but we are a bit busy with other things right now. I'll get back to you when I can find some room.

PeterMeisrimelModelon:
Do you happen to have a better reproducer of the original issue you are trying to fix here? While the added tests here would certainly traverse the added lines, the tests pass on the current master.

jschueller:
hello @PeterMeisrimelModelon, would you mind having a look at the updated test ? this time it exercises the fix correctly

PR #413, opened 2026-05-28
perf: ensure partial cache can be re-used also when stop_index is None

efredriksson-modelon:
This MR is to speedup when using get_variable_data method without any pagination. The main speedup is to not have to re-read the time vectors for each call.

### Motivation
The original guard rejected all stop_index=None cache hits. That was overly broad: it correctly blocked a partial cache being returned as a full trajectory, but it also unnecessarily blocked a genuine full trajectory in cache from being reused. The fix narrows the rejection to only the case that actually matters — cached length doesn't match the total number of available points — which is the precise condition under which returning the cache would truncate the result.

### DIFF
--- a/src/common/io.py
+++ b/src/common/io.py
@@ -1439,12 +1439,18 @@ class ResultDymolaBinary(ResultDymola):
         
         return name_dict
 
-    def _can_use_partial_cache(self, partial_trajectory: Union[np.ndarray, None], 
-                               stop_index: Union[int, None]) -> bool:
+    def _can_use_partial_cache(self, partial_trajectory: Union[np.ndarray, None],
+                               stop_index: Union[int, None],
+                               total_points: Union[int, None] = None) -> bool:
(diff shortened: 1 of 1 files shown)


src/common/io.py
```diff
@@ -1439,25 +1439,32 @@ def _get_name_dict(self):
...
-        # no file updates = full read only
-        # otherwise check sufficient length is available
-        return (partial_trajectory is not None) and (not self._allow_file_updates or (stop_index is not None and len(partial_trajectory) >= stop_index))
+        if partial_trajectory is None:
+            return False
+        if not self._allow_file_updates:
+            return True
+        if stop_index is None:
+            # A partial read stores fewer points than the full trajectory, so reusing it
+            # for an unbounded request would silently truncate the result.
```
PeterMeisrimelModelon:
```suggestion
            # Unbounded reads can re-use cache if there were no updates
```

PeterMeisrimelModelon (approved):

PR #414, opened 2026-07-10
feat: basic support for loading FMI3 CS FMUs.

PeterMeisrimelModelon:


src/pyfmi/fmi3.pyx
```diff
@@ -660,14 +754,14 @@ cdef class FMUModelBase3(FMI_BASE.ModelBase):
...
-    ):
-        raise NotImplementedError # to implemented in FMUModel(ME|CS|SE)3
+    # def initialize(self,
+    #     tolerance_defined=True,
+    #     tolerance="Default",
+    #     start_time="Default",
+    #     stop_time_defined=False,
+    #     stop_time="Default"
+    # ):
+    #     raise NotImplementedError # to implemented in FMUModel(ME|CS|SE)3
```
efredriksson-modelon: remove

tests/test_fmi3.py
```diff
@@ -1538,9 +1540,138 @@ def test_set_enum_case_sensitivity(self):
...
+    "Stair",
+    "VanDerPol",
+]
+
+class Test_FMI3CS:
+    """Basic unit tests for FMI3 import directly via the FMUModelCS3 class."""
+    def test_basic(self):
+        """Basic construction of FMUModelCS3."""
+        fmu = FMUModelCS3(FMI3_REF_FMU_PATH / "VanDerPol.fmu")
+        assert isinstance(fmu, FMUModelCS3)
```
efredriksson-modelon: I think this test can be removed. Feels like a weaker version of test_instantiation. The `isinstance` I think only makes sense when using `load_fmu` otherwise the class used is just what is used line above.

tests/test_fmi3.py
```diff
@@ -1538,9 +1540,138 @@ def test_set_enum_case_sensitivity(self):
...
+    def test_basic_wrong_fmu_type(self):
+        """Test using a non-CS FMU."""
+        fmu_path = str(this_dir / "files" / "FMUs" / "XML" / "ME3.0" / "variableAttributes")
+        msg = "The FMU could not be loaded. This class only supports FMI 3.0 for Co Simulation."
+        with pytest.raises(InvalidVersionException, match = msg):
+            FMUModelCS3(fmu_path, allow_unzipped_fmu = True, _connect_dll = False)
+
+    def test_get_model_identifier(self):
+        """Test that model identifier is retrieved as expected."""
+        fmu = FMUModelCS3(FMI3_REF_FMU_PATH / "VanDerPol.fmu")
```
efredriksson-modelon: This line is repeated in a lot of the tests so making this a fixture would make sense.

efredriksson-modelon (changes requested):
👍 

The old commented `def initialize` should definitely be removed. The test suggestions are minor optionals.

tests/test_fmi3.py
PeterMeisrimelModelon: Added some parametrizations here, see latest commit.

efredriksson-modelon (approved):
🍉

PR #415, opened 2026-07-10
feat: basic FMI3 CS doStep implementation

PeterMeisrimelModelon:
Not really re-viewable commit-by-commit.

Test re-structuring:

1. Moved the relevant new fixtures for FMI3 CS testing to `conftest.py` and the relevant helper functions to the new `utils.py`.

Comments for the actual changes:
1. My take of the FMI3 spec is that "terminate" is not formally included in the "earlyReturn", which I base on the https://fmi-standard.org/docs/3.0.2/#_basic_co_simulation in the specification. I.e., FMUs are allowed to return with `terminateSimulation = True`, even if instantiated with `allowEarlyReturn=False`
2. `terminate` now being part of the `do_step` signature is a breaking change, so we need a new mechanism to handle this. I think the 3 possible ways are a class attribute (the current variant), a helper function that fulfills the same purpose as the attribute, or make `do_step` have multiple returns. Multiple returns might be most aligned with the underlying C-API behavior, but least consistent with FMI2, but then again, this is a breaking change. Looking for input/suggestions here.

### DIFF
--- a/src/pyfmi/fmi3.pyx
+++ b/src/pyfmi/fmi3.pyx
@@ -3861,6 +3861,7 @@ cdef class FMUModelCS3(FMUModelBase3):
         if status != FMIL.jm_status_success:
             raise FMUException('Failed to instantiate the model. See the log for possibly more information.')
 
+        self._instantiated_with_early_return = earlyReturnAllowed
         self._allocated_fmu = 1
 
     cpdef _get_time(self):
@@ -3909,12 +3910,10 @@ cdef class FMUModelCS3(FMUModelBase3):
 
             status --
                     The status of function which can be checked against
-                    FMI_OK, FMI_WARNING. FMI_DISCARD, FMI_ERROR,
-                    FMI_FATAL,FMI_PENDING...
+                    FMI_OK, FMI_WARNING. FMI_DISCARD, FMI_ERROR, FMI_FATAL
 
         Calls the underlying low-level function fmi3DoStep.
         """
-        # TODO: Status docstring
         cdef FMIL3.fmi3_status_t status
         cdef FMIL3.fmi3_boolean_t new_s
         cdef FMIL3.fmi3_boolean_t eventHandlingNeeded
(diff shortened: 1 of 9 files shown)


src/pyfmi/fmi3.pyx
```diff
@@ -3888,6 +3895,171 @@ cdef class FMUModelCS3(FMUModelBase3):
...
+
+            new_step --
+                True the last step was accepted by the master and
+                False if not.
+
+        Returns::
+
+            status --
+                    The status of function which can be checked against
+                    FMI_OK, FMI_WARNING. FMI_DISCARD, FMI_ERROR, FMI_FATAL
```
efredriksson-modelon: `.` -> `,`

src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -1045,6 +1045,15 @@ def solve(self):
...
                 self.status = status
 
+                if isinstance(self.model, FMUModelCS3):
+                    if self.model.do_step_terminated:
+                        final_time = self.model.time
+
+                        start_time_point = timer()
+                        result_handler.integration_point()
+                        self.timings["storing_result"] += timer() - start_time_point
+                        break
```
efredriksson-modelon: This code is duplicated below, I think it would be good to break it out to a local function or something so these 2 sections does not drift in the future.

efredriksson-modelon (changes requested):
👍

src/pyfmi/fmi_algorithm_drivers.py
PeterMeisrimelModelon: Re-factored it a bit and it came out nicely, see latest commit. Also changed the order a bit to better handle error returns from `do_step`.

src/pyfmi/fmi_algorithm_drivers.py
```diff
@@ -1045,37 +1065,16 @@ def solve(self):
                 status = self.model.do_step(t,h)
                 self.status = status
 
-                if isinstance(self.model, FMUModelCS3):
-                    if self.model.do_step_terminated:
-                        final_time = self.model.time
+                _terminated, _terminated_time = self._check_do_step_status_and_terminated(status)
```
efredriksson-modelon: Why leading `_` for the names? It looks very un-Python/strange.

efredriksson-modelon (approved):
🍉 

(but please consider tweaking the variable names 😄)

PR #416, opened 2026-07-13
Adding support for CS output derivatives

PeterMeisrimelModelon:
There are no real tests, since none of the reference FMUs actually support this.

### DIFF
--- a/src/pyfmi/fmi3.pyx
+++ b/src/pyfmi/fmi3.pyx
@@ -3998,21 +3998,14 @@ cdef class FMUModelCS3(FMUModelBase3):
         if order < 1 or order > max_output_derivative:
             raise FMUException("The order must be greater than zero and below the maximum output derivative support of the FMU (%d)."%max_output_derivative)
 
-        if isinstance(variables, str):
-            nref = 1
-            value_refs = np.array([0], dtype=np.uint32, ndmin=1).ravel()
-            orders = np.array([order], dtype=np.int32)
-            value_refs[0] = self.get_variable_valueref(variables)
-        elif isinstance(variables, list) and np.prod([int(isinstance(v, str)) for v in variables]): #prod equals 0 or 1
-            nref = len(variables)
-            value_refs = np.array([0]*nref, dtype=np.uint32, ndmin=1).ravel()
-            orders = np.array([0]*nref, dtype=np.int32)
-            for i in range(nref):
-                value_refs[i] = self.get_variable_valueref(variables[i])
-                orders[i] = order
-        else:
-            raise FMUException("The variables must either be a string or a list of strings")
+        if not isinstance(variables, (str, list)) or not all(isinstance(v, str) for v in variables):
+            raise FMUException("The variables must either be a string or a list of strings.")
 
+        if isinstance(variables, str):
+            variables = [variables]
+        nref = len(variables)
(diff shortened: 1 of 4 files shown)


src/pyfmi/fmi3.pyx
```diff
@@ -3968,6 +3968,77 @@ cdef class FMUModelCS3(FMUModelBase3):
...
+
+        if order < 1 or order > max_output_derivative:
+            raise FMUException("The order must be greater than zero and below the maximum output derivative support of the FMU (%d)."%max_output_derivative)
+
+        if isinstance(variables, str):
+            nref = 1
+            value_refs = np.array([0], dtype=np.uint32, ndmin=1).ravel()
+            orders = np.array([order], dtype=np.int32)
+            value_refs[0] = self.get_variable_valueref(variables)
+        elif isinstance(variables, list) and np.prod([int(isinstance(v, str)) for v in variables]): #prod equals 0 or 1
```
efredriksson-modelon: Could we not just use `all(isinstance(v, str)) for v in variables)` here?

src/pyfmi/fmi3.pyx
```diff
@@ -3968,6 +3968,77 @@ cdef class FMUModelCS3(FMUModelBase3):
...
+
+        max_output_derivative = FMIL3.fmi3_import_get_capability(self._fmu, FMIL3.fmi3_cs_maxOutputDerivativeOrder)
+
+        if order < 1 or order > max_output_derivative:
+            raise FMUException("The order must be greater than zero and below the maximum output derivative support of the FMU (%d)."%max_output_derivative)
+
+        if isinstance(variables, str):
+            nref = 1
+            value_refs = np.array([0], dtype=np.uint32, ndmin=1).ravel()
+            orders = np.array([order], dtype=np.int32)
```
efredriksson-modelon: To have less repetition we can use:
```
if isinstance(variables, str):
  variables = [variables]
```

to have this case transform to the general

src/pyfmi/fmi3.pyx
```diff
@@ -3968,6 +3968,77 @@ cdef class FMUModelCS3(FMUModelBase3):
...
+            value_refs[0] = self.get_variable_valueref(variables)
+        elif isinstance(variables, list) and np.prod([int(isinstance(v, str)) for v in variables]): #prod equals 0 or 1
+            nref = len(variables)
+            value_refs = np.array([0]*nref, dtype=np.uint32, ndmin=1).ravel()
+            orders = np.array([0]*nref, dtype=np.int32)
+            for i in range(nref):
+                value_refs[i] = self.get_variable_valueref(variables[i])
+                orders[i] = order
+        else:
+            raise FMUException("The variables must either be a string or a list of strings")
```
efredriksson-modelon: Consider having the error check first for an early return to have less cognitive load/indentation when reading the actual implementation.

efredriksson-modelon (changes requested):
Looks good 🍎, some suggestions for making the implementation easier to read.

src/pyfmi/fmi3.pyx
PeterMeisrimelModelon: good suggestions; cleaned it up a bit, a lot nicer to read.

efredriksson-modelon (approved):
🍉

PR #419, opened 2026-08-31
fix: fixed an exception when using dynamic_diagnostics for states with unbounded=true

PeterMeisrimelModelon:


efredriksson-modelon (approved):
Nice way to solve it 🍉

PR #420, opened 2026-09-07
Adding get_fmu_kind function

PeterMeisrimelModelon:


src/pyfmi/fmi.pyx
```diff
@@ -243,7 +244,10 @@ cpdef load_fmu(fmu: Union[str, Path], log_file_name = None, kind = 'auto',
     fmu_full_path = os.path.abspath(fmu)
     check_fmu_args(allow_unzipped_fmu, fmu, fmu_full_path)
 
-    # Check that kind-argument is well-defined
+    # Check that kind-argument is well-defined.
+    # Note that 'kind' is normalized into an exact str, since the version specific
+    # loaders below are typed 'str kind'
+    kind = str(kind)
     _allowed_kinds = ["ME", "CS", "SE"]
```
efredriksson-modelon: As we are only using this to validate input with .upper we could consider changing these to lower case and validate with 'lower' and also change the docstring to have ME -> me, CS -> cs just to be consistent. It would not really change behavior as we accept any capitalization.

src/pyfmi/fmi1.pyx
```diff
@@ -446,6 +449,8 @@ cdef class FMUModelBase(FMI_BASE.ModelBase):
             model.get_version()
         """
         cdef FMIL1.fmi1_string_t version
+        if not self._allocated_dll:
+            return "1.0" # only valid value according to specification
```
efredriksson-modelon: I think we should remove this as not having special case for if DLL is not loaded.

efredriksson-modelon (changes requested):
🍉

src/pyfmi/fmi.pyx
PeterMeisrimelModelon: Fixed. Also added explicit tests that check that both upper and lower case variants are accepted.

src/pyfmi/fmi1.pyx
PeterMeisrimelModelon: Fixed.

efredriksson-modelon (approved):
👍

